PR #51, opened 2013-07-10
Feature/hof

tmaus:
HoF and Rankings is now available. 

Pls. run first checks


hemberger:
I think I understand the Operations ranking that are implemented here, but I don't see what the "Rank Management" stuff is doing. Is it worth me resolving the conflicts and getting this into a mergable state, or should this PR be closed?

PR #57, opened 2014-02-14
Research

tmaus:
Research features allows alliance to research ships on their respective planets.


hemberger:
This is very interesting, but I think it's a bit too ancient to be resurrected at this point (especially since the original author is no longer active).

PR #64, opened 2014-07-08
Working on Modals

wcyrek:
Added some modal JS
Changed default bootstrap CSS
Added more class tags to galaxy map


hemberger:
For the sake of cleaning up some of the old PR's, should this remain open? From what @wcyrek has said, it's not clear to me what this does or if it's working. Any updates for 3 years later? :)

Issue #15, opened 2012-08-03
Confirmation On Planet Bonding


Page-:
Javascript confirmation alert?

PR #71, opened 2016-07-02
Issue #15. Warning dialog for planet bonding.

Curufir:
Still need to figure out a way to style the dialog box. I've been playing with doing it dynamically using JQuery but I think it would be better to use the JQuery themeroller instead. As it stands the jquery-ui CSS overrides pretty much any changes.


Page-:
The indentation is a bit wonky (it seems to have a bunch of mixed tabs + spaces - generally I think the files are all tabs), but otherwise it looks good - if you can fix the indentation I'll merge straight away

Curufir:
Still getting used to Emacs. The auto-indenting seems to freak out on mixed files. Fixed now.

Page-:
Thanks :)

Issue #67, opened 2014-10-27
CSRF state token does not match one provided


MichaelKunze:
There might be a problem with the login_processing.php and Facebook.

I'm getting alot of those in my server log:

2014/10/27 04:03:31 [error] 17076#0: *4093313 FastCGI sent in stderr: "PHP message: CSRF state token does not match one provided" while reading response header from upstream, client: xxx.xxx.xxx.xxx, server: www.smrealms.de, request: "GET /login_processing.php?loginType=Facebook&code=XXXXXXX&state=XXXXXX HTTP/1.1", upstream: "fastcgi://unix:/some/path/php5-fpm.sock:", host: "www.smrealms.de", referrer: "https://www.facebook.com/dialog/oauth?client_id=XXXXX&redirect_uri=https%3A%2F%2Fwww.smrealms.de%2Flogin_processing.php%3FloginType%3DFacebook&state=XXXXX&scope=email"

hemberger:
As of bb09c388de98e9c5cc384444c1011586342515f2, we install the Facebook SDK with composer, rather than including the source code directly. This should avoid Facebook login-related errors, but please re-open if it becomes an issue again.

PR #72, opened 2016-08-01
Issue #67 - Updated Facebook SDK to v5.2

Curufir:


lib/Login/SocialLogin.class.inc
```diff
@@ -1,4 +1,35 @@
 <?php
+// Autoloader for the new Facebook API
+require_once(LIB . 'External/Facebook/autoload.php');
```
Page-: This would be better in the `getFacebookObj` function as it was before, it avoids the need for loading/using the facebook classes when it's not need or is disabled due to not having facebook configs set (also as a general thing it keeps the behaviour as similar as possible, to avoid weird things)

lib/Login/SocialLogin.class.inc
```diff
@@ -1,4 +1,35 @@
...
+
+// It would be cleaner to store the facebook CSRF token in the session but this will do.
+if(!session_id()) {
```
Page-: I see some very similar code to this in the twitter code - it's probably best to consolidate the code between the two

lib/Login/SocialLogin.class.inc
```diff
@@ -1,4 +1,35 @@
...
+	session_start();
+}
+
+
+// Hackish but it's because we're not on v5.6 of PHP which has this builtin
+if(!function_exists('hash_equals'))
```
Page-: I see this is also in `lib/External/Facebook/polyfills.php`, can we just load that file instead?

PR #88, opened 2017-07-21
Refactor information in planet_type table

hemberger:
With some requested changes to Arid Planets in the pipeline, I felt it would be easier to refactor the planet_type information instead of continuing to deal with database patches for static data. The first commit simply removes two unused and redundant functions, but the second does the actual refactoring. See the commit messages for more detail.


hemberger:
Page has suggested that we might want to provide an admin interface to modify these tables rather than hard-code them into the php. I still personally prefer making changes to the code, since those can be tracked and audited much more easily than changes to the database, but for now it looks like this type of change will require ongoing discussion. Closing this PR for now.

PR #93, opened 2017-07-22
Add credits lost message to deaths from ports/planets/forces

hemberger:
This patch provides extra information about how many credits are lost when a player is destroyed by ports, planets, and forces. I don't believe this information is available to the destroyed player in any other form, which seems like an oversight.

The message reads:
"The X credits that were onboard Y's ship are lost in the wreckage."


hemberger:
Here is what it looks like in context.
image 1

PR #95, opened 2017-07-26
Minor display changes to 1.6/universe_create.php

hemberger:
Adds some more information to the initial 1.6 universe creation screen (headers and game numbers in the edit game dropdown). Moves the "Edit Existing Games" form to the top of the page, since it is easy to overlook otherwise, due to the much larger "Create New Game" form.


hemberger:
image

PR #99, opened 2017-07-27
Fix ajax updates

hemberger:
These commits make ajax updates more robust in general by ensuring that `xmlify` produces valid XML, and in the process fixes a specific issue with the Planet Construction page.

### DIFF
--- a/templates/Default/engine/Default/planet_construction.php
+++ b/templates/Default/engine/Default/planet_construction.php
@@ -1,17 +1,18 @@
-<p>You are currently building: <?php
+You are currently building: <?php
 	if ($ThisPlanet->hasCurrentlyBuilding()) {
 		$CurrentlyBuilding = $ThisPlanet->getCurrentlyBuilding();
 		foreach($CurrentlyBuilding as $Building) { ?>
 			<br /><?php
 			echo $PlanetBuildings[$Building['ConstructionID']]['Name']; ?> which will finish in <?php echo format_time($Building['TimeRemaining']); ?>
+			<br /><br />
 			<div class="buttonA"><a class="buttonA" href="<?php echo $ThisPlanet->getCancelHREF($Building); ?>">&nbsp;Cancel&nbsp;</a></div><?php
 		}
 	}
 	else { ?>
 		Nothing!<?php
 	} ?>
-</p>
 
+<br /><br />
 <table class="standard">
 	<tr>
 		<th></th>
--- a/engine/Default/smr.inc
+++ b/engine/Default/smr.inc
@@ -124,7 +124,7 @@ function smrBBCode($bbParser, $action, $tagName, $default, $tagParams, $tagConte
 }
 
 function &xmlify(&$str) {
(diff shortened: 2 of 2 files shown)


hemberger:
Here is an example of the issue with the DOM parsing this PR fixes.

This is the raw html:
```
<p>You are currently building: <br />Generator which will finish in 1 hour, 20 minutes and 9 seconds <div class="buttonA"><a class="buttonA" href="?sn=f0f7f295">&nbsp;Cancel&nbsp;</a></div></p>
```

This is the output from `phpQuery->html()`:
```
<p>You are currently building: <br>Generator which will finish in 1 hour, 18 minutes and 37 seconds </p>\n<div class="buttonA"><a class="buttonA" href="?sn=2b65ab93">\xc2\xa0Cancel\xc2\xa0</a></div>\n
```

hemberger:
For reference, [xml_comparison.zip](https://github.com/smrealms/smr/files/1181599/xml_comparison.zip) contains the XML files for the Planet Construction page with both the old and new XML escaping.

templates/Default/engine/Default/planet_construction.php
```diff
@@ -1,4 +1,4 @@
-<p>You are currently building: <?php
+<p><span name="currently_building">You are currently building: <?php
```
Page-: Can you remove this span (or fully convert the file to specific refresh areas) - as it stands it just adds more html without changing anything (at least I don't think it changes anything)

hemberger: The added `<span>` does not introduce a refresh area. The entire middle panel is being updated even without this change (which I believe is desirable and the original intent for this page).

What this change _does_ do is fix a weird issue with the "Cancel" button jumping to a different location on the page after an ajax update. Specifically, the DOM parsing of that one particular `<p>...</p>` on this page is broken. As you can see from https://github.com/smrealms/smr/pull/99#issuecomment-318519929, the `convertHtmlToAjaxXml` code that is only run to determine how the page looks after an ajax update puts the `</p>` in the wrong place! Adding the `<span>` forces it to keep the `</p>` in the right place. (If you'd like to see it visually, I can take screenshots with and without the span.)

I don't understand why the DOM parsing is buggy, but it is. This change only circumvents the bug, so if you have some idea how to fix it without introducing a `<span>`, I'm all for it! But if it's just a choice between correct ajax updates and an extra html element vs. incorrect ajax updates, I'd take the former any day. :)

hemberger: I also tested the update to phpQuery in #107, but it did not fix this bug.

Page-: I expect the issue is that because a `<div>` is not a valid element to be contained within a `<p>` the html parser automatically ends the `<p>` when it sees the `<div>`.  It probably makes sense to drop the `<p>` altogether

hemberger:
I have modified the second commit (fixing `planet_construction.php` template) to address the underlying problem identified by Page. Instead of adding a `<span>` to circumvent the problem, we simply remove the illegal `<p>` (because a `<p>` is not allowed to contain a `<div>`). This causes a minimal change to the look of the page.

Before:
image

After:
image

PR #101, opened 2017-07-28
shop_goods.php: fix broken form html

hemberger:
Replace a `<form>` that was supposed to be `</form>`.


hemberger:
Old:
```
<form name="FORM" method="POST" action="?sn=2aa7302f">
<input type="submit" name="action" value="Leave Port" id="InputFields" />
<form>
```
New:
```
<form name="FORM" method="POST" action="?sn=5f6a3f8b">
<input type="submit" name="action" value="Leave Port" id="InputFields" />
</form>
```

PR #105, opened 2017-07-29
New admin tools

hemberger:
Admins can now be given permission to do the following:
* Enable new games
* Assign or remove Galactic Post editors

A flyway migrate must be run for these options to take effect.

### DIFF
--- a/admin/Default/enable_game.php
+++ b/admin/Default/enable_game.php
@@ -15,7 +15,7 @@ while ($db->nextRecord()) {
 $template->assign('DisabledGames', $disabledGames);
 
 // Create the link to the processing file
-$link_container = create_container('enable_game_processing.php', '');
-$template->assign('EnableGameHREF', SmrSession::getNewHREF($link_container));
+$linkContainer = create_container('enable_game_processing.php', '');
+$template->assign('EnableGameHREF', SmrSession::getNewHREF($linkContainer));
 
 ?>
--- a/admin/Default/enable_game.php
+++ b/admin/Default/enable_game.php
@@ -1,0 +1,21 @@
+<?php
+
+$template->assign('PageTopic', 'Enable New Games');
+
+// If we have just forwarded from the processing file, pass its message.
+$template->assign('ProcessingMsg', $var['processing_msg']);
+
+// Get the list of disabled games
+$db->query('SELECT game_name, game_id FROM game WHERE enabled=' . $db->escapeBoolean(false));
+$disabledGames = array();
+while ($db->nextRecord()) {
+	$disabledGames[] = array('game_name' => $db->getField('game_name'),
+	                         'game_id' => $db->getInt('game_id'));
+}
+$template->assign('DisabledGames', $disabledGames);
+
+// Create the link to the processing file
+$link_container = create_container('enable_game_processing.php', '');
+$template->assign('EnableGameHREF', SmrSession::getNewHREF($link_container));
+
+?>
--- a/admin/Default/enable_game_processing.php
+++ b/admin/Default/enable_game_processing.php
@@ -1,0 +1,16 @@
+<?php
+
+if (!empty($_POST['game_id'])) {
+	// Enable the requested game
+	$game_id = $db->escapeNumber($_POST['game_id']);
+	$enabled = $db->escapeBoolean(true);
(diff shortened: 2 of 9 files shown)


hemberger:
image
image

hemberger:
image
image

admin/Default/enable_game_processing.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+if (!empty($_POST['game_id'])) {
+	// Enable the requested game
+	$game_id = $db->escapeNumber($_POST['game_id']);
```
Page-: Can you put the escaping inline with the query - it just helps to make it obvious that the escaping has been done (I mean the ideal would be to use bind variables everywhere, but that'd be quite a huge change :( )

admin/Default/manage_post_editors.php
```diff
@@ -0,0 +1,44 @@
...
+while ($db->nextRecord()) {
+	$activeGames[] = array('game_name' => $db->getField('game_name'),
+	                       'game_id' => $db->getInt('game_id'));
+}
+$template->assign('ActiveGames', $activeGames);
+
+if ($activeGames) {
+	// Set the selected game (or the first in the list if not selected yet)
+	if (isset($_POST['game_id'])) {
+		SmrSession::updateVar('selected_game_id', $_POST['game_id']);
```
Page-: It would be nice to check that the chosen game id is a valid number and display a nice message, right now it will throw an error at `$db->escapeNumber($var['selected_game_id'])` instead which will be a bit ugly - at the same time I'm not too fussed as this is in admin tools

admin/Default/enable_game.php
```diff
@@ -0,0 +1,21 @@
...
+$db->query('SELECT game_name, game_id FROM game WHERE enabled=' . $db->escapeBoolean(false));
+$disabledGames = array();
+while ($db->nextRecord()) {
+	$disabledGames[] = array('game_name' => $db->getField('game_name'),
+	                         'game_id' => $db->getInt('game_id'));
+}
+$template->assign('DisabledGames', $disabledGames);
+
+// Create the link to the processing file
+$link_container = create_container('enable_game_processing.php', '');
```
Page-: Can you use `camelCase` variables please - just to try to keep it standardised throughout the code base as much as possible (although with php that's never going to be fully possible as it has multiple styles for built in stuff as it is :(  )

templates/Default/admin/Default/enable_game.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+// This var is passed by the processing file if we enabled a game
+if (!empty($ProcessingMsg)) {
+	echo $ProcessingMsg;
+}
+
+if (empty($DisabledGames)) {
+	echo "<p>All games are already enabled!</p>";
```
Page-: Can you do this via
```
?><p>All games are already enabled!</p><?php
```
instead please, it lets the html easily be syntax highlighted/validated by editors (you can see github has highlighted the html below due to being outside of the php tags)

admin/Default/enable_game_processing.php
hemberger: Fixed in this place and another I found.

admin/Default/enable_game.php
hemberger: Fixed in this place and others.

templates/Default/admin/Default/enable_game.php
hemberger: Fixed.

admin/Default/manage_post_editors.php
hemberger: The list of possible `game_id`'s that they can select from in the dropdown menu is populated by a database query (`$activeGames`), so in that sense `$_POST['game_id']` must provide a valid game ID. If they manipulate the form to be able to submit a game ID that was not provided... well then perhaps they deserve to get a database error. ;)

Page-: The problem is that if instead of a db error they get access to something they shouldn't :P  In this case it doesn't change anything as it's possible to change the GP editor of any game, but similar code for the case of map editing would allow any admin to edit any game, even without the "Edit Live Games" privilege - it's just worth being aware of the potential pitfalls for the cases where it really does matter

hemberger: Understood. Thanks for the advice!

PR #113, opened 2017-08-03
Switch to mysqli database connector

hemberger:
The mysql connector is deprecated in php 5.5 and removed from
php 7+. This change will allow us to seamlessly upgrade to php 7.

### DIFF
--- a/htdocs/config.inc
+++ b/htdocs/config.inc
@@ -13,9 +13,11 @@ function logException(Exception $e) {
 			'E-Mail: '.$account->getEmail().EOL.EOL.'-----------'.EOL.EOL;
 	}
 	$message .= 'Error Message: '.$e->getMessage().EOL.EOL.'-----------'.EOL.EOL;
-	if($currMySQLError = mysql_error()) {
+	require_once(LIB . '/Default/SmrMySqlDatabase.class.inc');
+	$db = new SmrMySqlDatabase();
+	if($currMySQLError = $db->getError()) {
 		$errorType = 'Database Error';
-		$message .= 'MySQL Error MSG: '.mysql_error().EOL.EOL.'-----------'.EOL.EOL;
+		$message .= 'MySQL Error MSG: '.$db->getError().EOL.EOL.'-----------'.EOL.EOL;
 	}
 	$message .=	'Trace MSG: '.$e->getTraceAsString().EOL.EOL.'-----------'.EOL.EOL.
 		'$var: '.var_export($var,true).EOL.EOL.'-----------'.EOL.EOL.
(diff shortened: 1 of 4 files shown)


htdocs/config.inc
```diff
@@ -13,9 +13,11 @@ function logException(Exception $e) {
 			'E-Mail: '.$account->getEmail().EOL.EOL.'-----------'.EOL.EOL;
 	}
 	$message .= 'Error Message: '.$e->getMessage().EOL.EOL.'-----------'.EOL.EOL;
-	if($currMySQLError = mysql_error()) {
+	require_once(LIB . '/Default/SmrMySqlDatabase.class.inc');
+	$db = new SmrMySqlDatabase();
```
Page-: Right now this potentially creates a db instance/connection when we're in an error case, which I'd rather avoid, I think this can probably be done via `global $db` and checking if there's already one available, if not's not globally available usually then using `class_exists` should do the trick.  I doubt it's the end of the world if we do, but it'd be preferable not to - if we do have to then we should at least try/catch it so we can still send the original message, like with `release_lock()` below

lib/Default/MySqlDatabase.class.inc
```diff
@@ -3,39 +3,33 @@ require_once('Database.interface.inc');
...
 	protected $dbRecord = null;
 	protected $debug = true;
 	
 	public function __construct($host, $user, $password, $databaseName) {
 		if(!self::$dbConn) {
-			if(!self::$dbConn = @mysql_connect($host, $user, $password)) {
-				$this->error('Connection failed.');
+			self::$dbConn = new mysqli($host, $user, $password, $databaseName);
+			if (self::$dbConn->connect_errno) {
+				$this->error('Connection failed: ' . self::$dbConn->connect_error);
```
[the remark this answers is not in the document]
hemberger: Exception messages are only presented to the user when `ENABLE_DEBUG` is set, so this is already hidden behind a debug flag. When debug is off, exception messages are e-mailed to bugs@smrealms.de (yes, that should not be hard-coded, but the point is that it's private :P)

htdocs/config.inc
hemberger: Based on http://php.net/manual/en/mysqli.construct.php, it looks like the `mysqli` constructor is exception-free. Since an object is returned even if `mysqli(...)` fails, then `self::$dbConn (a class static) will be set and future calls to the `MySqlDatabase` constructor will succeed (because it will see that `self::$dbConn` exists and therefore skip the failing `mysqli` call. This allows us to get the error message as we expect.

On the other hand, if we were to use `global $db`, then we will actually miss the error message because the first (failing) call to the `MySqlDatabase` constructor never completes, and so `$db` won't exist in `logException`.

I believe the code is correct as-is, unless there is some case that you think I've forgotten.

PR #115, opened 2017-08-14
Use docker-compose to manage services

hemberger:
This patch simplifies the service management by using `docker-compose`. All services (smr, smr-dev, mysql, phpmyadmin, and flyway) are maintained in the `docker-compose.yml` file. The updated management commands are reflected in the changes to `README.md`, which are much shorter and require less information to manually specified.

A refactor of `Smr{,Session}MySqlDatabase.class.inc` followed by the inclusion of specifying mysql passwords in an uncommitted `.env` file allow us to avoid needing to link any of these files. The `.env` file also specifies the path to the `config.specific.php` file.

See the commit messages for details about each change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# Dockerfile for smr service
 FROM php:5.6-apache
 RUN apt-get update \
 	&& apt-get install -y libcurl4-openssl-dev git sendmail \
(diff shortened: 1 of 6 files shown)


hemberger:
I'm going to split this up into multiple PR's.

PR #116, opened 2017-08-15
combat_log_viewer.php: include type of logs in message

hemberger:
Before this file was converted to a template, the combat log viewer
would identify the type of logs in the message (e.g. "There are 4
alliance logs available..."). In the template conversion, the `LogType`
template variable was set, but appears to have been mistakenly unused.
We could have simply removed this variable, but the extra descriptive
message is helpful (even if only to identify what page we're on),
so we add the LogType back in.


hemberger:
The updated messages look like:

image

image

PR #117, opened 2017-08-16
Add BBLink for combat logs

hemberger:
Use BBLinks as a secure way to provide links to combat logs in messages sent to players. We implement a new case for the BBCode parser `combatlogs`, and then use this now in planet attack messages. This can be used for other messages in the future (e.g. player death, player kills, force attacks) with some additional work.

For the links, we use the image that used to be for reporting messages to admins, and replace that with an in-theme red exclamation point image.

See the commit messages for more details.


hemberger:
The modified messages look as follows (notice the added link and the altered "report to admin" button):

image

engine/Default/smr.inc
```diff
@@ -23,6 +23,17 @@ function smrBBCode($bbParser, $action, $tagName, $default, $tagParams, $tagConte
 	global $overrideGameID, $disableBBLinks, $player, $account, $var;
 	try {
 		switch($tagName) {
+			case 'combatlog':
+				$logID = (is_numeric($default)) ? $default : $tagParams['id'];
+				if ($action == BBCODE_CHECK) {
+					return true;
+				}
+				$container = create_container('skeleton.php', 'combat_log_viewer.php');
+				$container['action'] = 5;
```
Page-: I'd love to see these as named constants because the numbers are not the easiest to associate with actions (although I know 5 is 'view')

hemberger: I agree, and I'd be happy to make this change. Where would you suggest I define the constants?

Page-: https://github.com/smrealms/smr/blob/master/htdocs/config.php probably makes sense

PR #121, opened 2017-08-17
Improved php debugging

hemberger:
When in debug mode (as defined by `ENABLE_DEBUG` in `config.specific.php`), all php notices will be printed to help catch bugs. A quick pass through the code with these notices turned on revealed a few issues, which are fixed in this PR.

See commit messages for more details.

### DIFF
--- a/lib/Default/SmrPlanet.class.inc
+++ b/lib/Default/SmrPlanet.class.inc
@@ -43,6 +43,7 @@ class SmrPlanet {
 
 	protected $delayedShieldsDelta = 0;
 	protected $delayedCDsDelta = 0;
+	protected $delayedArmourDelta = 0;
 
 	public static function refreshCache() {
 		foreach(self::$CACHE_PLANETS as $gameID => &$gamePlanets) {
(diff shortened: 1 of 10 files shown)


lib/Default/SmrPlanet.class.inc
```diff
@@ -847,6 +848,8 @@ class SmrPlanet {
 		$this->delayedShieldsDelta = 0;
 		$this->setCDs($this->getCDs(true));
 		$this->delayedCDsDelta = 0;
+		$this->setArmour($this->getArmour(true));
```
Page-: I think this might be able to cause issues with when the armour hits 0, at the same time delayed armour is possible to be added, so it should be dealt with here and this just reduces the possible bugs (I don't think anywhere actually adds it though)

lib/Default/SmrSession.class.inc
```diff
@@ -300,7 +300,9 @@ class SmrSession {
 			$container['RemainingPageLoads'] = 1; // Allow refreshing
 		}
 		if(!isset($container['PreviousRequestTime'])) {
-			$container['PreviousRequestTime'] = self::$var[$sn]['PreviousRequestTime']; // Copy across the previous request time if not explicitly set.
+			if(isset(self::$var[$sn]['PreviousRequestTime'])) {
```
Page-: This could be merged into the `if` above

PR #122, opened 2017-08-18
SmrPlayer.class.inc: fix planet message whitespace

hemberger:
In the message sent to planet owners when a planet kills an attacker,
it used a regular space between two HTML tags, which gets removed,
which results in two words being merged into one. We add a `&nbsp;`
to ensure that a space is added between the tags.


hemberger:
Bottom message: before this patch
Top message: after this patch
image

PR #123, opened 2017-08-19
SmrPlayer.class.inc: include force owner in death message

hemberger:
Even though the force owner is displayed as the sender of the
message, we add the force owner in the text for consistency.
This is an extension of PR #69, which added the force owner to
the news entry.


hemberger:
Top: new message
Bottom: old message
image

PR #127, opened 2017-08-24
1.6/universe_create.php: handle empty existing games

hemberger:
If there were no valid games to edit, the existing games dropdown
would be empty and clicking the "Edit" button would cause a mysql
error.

To avoid this, we omit the dropdown if there are no valid games
to edit, replacing it with some explanatory text.


hemberger:
An example of the new text:
image

PR #128, opened 2017-09-01
Fix logging of sector movement

hemberger:
Sector movement was incorrectly logged because the target sector
and the current sector were set to the same value. This bug was
introduced in 30a7213e8b16, when logging for sector movement was
wrapped inside the mission actions.

Since 'EnterSector' superseded 'WalkSector' as the mission action
to detect if a player was in a sector (since 'EnterSector' accounts
for jumping into a sector, for example), there is no reason to keep
'WalkSector' as a mission action, since its only purpose was for
logging. Instead, we simply log the movement directly as we did
before the bug was introduced.

### DIFF
--- a/engine/Default/sector_move_processing.php
+++ b/engine/Default/sector_move_processing.php
@@ -88,7 +88,8 @@ if ($player->hasPlottedCourse()) {
 }
 
 // log action
-$player->actionTaken('WalkSector',array('Sector'=>&$sector));
+$targetSector =& SmrSector::getSector($player->getGameID(), $var['target_sector']);
+$player->actionTaken('WalkSector', array('Sector' => &$targetSector));
 
 // send scout msg
 $sector->leavingSector($player,MOVEMENT_WALK);


hemberger:
Old behavior:
image

New behavior:
image

engine/Default/sector_move_processing.php
```diff
@@ -88,7 +88,7 @@
 }
 
 // log action
-$player->actionTaken('WalkSector',array('Sector'=>&$sector));
+$account->log(LOG_TYPE_MOVEMENT, 'Walks to sector: ' . $var['target_sector'], $player->getSectorID());
```
Page-: Can you change this to
```
$player->actionTaken('WalkSector',array('Sector'=>SmrSector::getSector($var['target_sector'])));
```
instead, that way it should still fix the bug, and still allows the ability for missions to distinguish between walking to a sector and getting there via other means

hemberger: I can make this change, but first I just want us to briefly consider if it's necessary. I would like to eventually add a mission for each race (so that Salvene isn't uniquely advantaged), but I can't think of any case where I'd want to penalize someone for jumping into a sector. If anything, it seems more likely to be used accidentally if a coder isn't aware of `EnterSector`.

I leave it up to you, though, so let me know. Thanks!

Page-: I think it's nice to have, in that it lets the logging be largely in a standardised place and it offers up the power of distinguishing between types of movement in missions, although just because it's there doesn't mean it has to be used (I think there's a fair few "missionable" actions right now that no missions use)

hemberger: Fair enough! I'll make this change shortly.

PR #129, opened 2017-09-01
Fix bug in saving MICRO_TIME to the database

hemberger:
The `microtime` built-in function outputs a string that looks like:

0.39324900 1504296581

The first word is the number of microseconds in seconds.
The function `microtimeMSec` took the 6 digits after the decimal
from the first word and converted that to an int.

However, if the number of microseconds was small enough to start
with a `0` after the decimal (e.g. `0.09324900`), then casting
the substring to an int would give `93249` instead of `093249`.
Therefore, when reconstructing the time from this value, we would
be one digit off, and you'd get a timestamp from 1974 or so.
[Thanks to Curufir for identifying this issue.]

By having `microtimeMSec` (and `microtimeSec`) return strings instead
of ints, we avoid the casting problem altogether.

We also move the sanity check on the microtime argument format
(i.e. that it has a length of 21 characters) out of `escapeMicrotime`,
which is agnostic to the format, and into `microtime{M,}Sec`, which
extract substrings that depend on the precise string format.

Due to the limited use of `escapeMicrotime`, this bug only affected
saving MICRO_TIME into the `account_has_logs` table. With this fix,
the Log Console now properly displays the event times.

-----------

**NOTE**: I have also added a backwards compatibility fix for the incorrectly stored `microtime`'s in the database. However, if you don't mind throwing out all the old logs, a cleaner fix to this bug is to simply avoid parsing `microtime()` output in the first place. Instead, we could use `microtime(true)`, which returns the microtime as a float instead of two strings (and then we would store it as a float in the database).

### DIFF
--- a/admin/Default/log_console_detail.php
+++ b/admin/Default/log_console_detail.php
@@ -169,7 +169,7 @@ else {
 	$db->query('SELECT * FROM account_has_logs WHERE account_id IN ('.$account_list.') AND log_type_id IN ('.$db->escapeArray($log_type_id_list).') ORDER BY ' . $var['item'] . ' ' . $var['order']);
 	while ($db->nextRecord()) {
 		$account_id		= $db->getInt('account_id');
-		$microtime		= $db->getMicrotime('microtime');
+		$microtime		= $db->getMicrotime('microtime', true); //fix value length errors
 		$message		= stripslashes($db->getField('message'));
 		$log_type_id	= $db->getInt('log_type_id');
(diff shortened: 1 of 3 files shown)


hemberger:
A few examples of the corrupted `microtime` field in the database:
image

hemberger:
Honestly, it would take me a while to get back into this to the point where I remembered how I wanted to do the error handling and database migration for a float microtime. Unless someone else has time to make the change on a short timescale, I'd just go with this PR because it's a tested fix.

PR #130, opened 2017-09-03
log_console_detail.php: add 2 more colors

hemberger:
Logs from different accounts are color-coded, but the color cycle
only has 4 colors. To avoid repeated colors if viewing more than
4 accounts, 2 additional colors (light blue and pink) were added.

The colors that are hardest to read on the dark green background
(red and blue) were moved to the end of the color cycle.

NOTE: Another option would be to include a column for the account
name instead of, or in addition to, the colors.

### DIFF
--- a/admin/Default/log_console_detail.php
+++ b/admin/Default/log_console_detail.php
@@ -37,7 +37,8 @@ else {
 	$db2 = new SmrMySqlDatabase();
 
 	// predefine a color array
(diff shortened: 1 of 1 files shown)


hemberger:
Before (with degenerate colors):
screenshot from 2017-09-02 22-59-39

After (with unique colors):
image

image

PR #133, opened 2017-09-04
Add ability to create planets of any type

hemberger:
The "Create Planets" page now has a list of all types of planets, rather than just creating Terran planets.

We also convert this page to a template and clean up some unused code.

image

### DIFF
--- a/admin/Default/1.6/universe_create_save_processing.php
+++ b/admin/Default/1.6/universe_create_save_processing.php
@@ -167,8 +167,6 @@ elseif ($submit == 'Create Planets') {
 	$galaxies =& SmrGalaxy::getGameGalaxies($var['game_id']);
 	$galaxy =& SmrGalaxy::getGalaxy($var['game_id'],$var['gal_on']);
 	$galSectors =& $galaxy->getSectors();
-	//get totals
-	$numberOfPlanets=0;
 	foreach ($galSectors as &$galSector) {
 		if($galSector->hasPlanet()) {
 			$galSector->removePlanet();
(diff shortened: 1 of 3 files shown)


templates/Default/admin/Default/1.6/universe_create_planets.php
```diff
@@ -0,0 +1,23 @@
+Working on Galaxy : <?php echo $Galaxy->getName(); ?> (<?php echo $Galaxy->getGalaxyID(); ?>)
+<br /><br />
+
+<?php echo $Form; ?>
```
Page-: Can you change this to
```php
$template->assign('CreatePlanetsFormHREF', SmrSession::getNewHREF($container));
...
<form method="POST" action="<?php echo $CreatePlanetsFormHREF; ?>">
```
so we can get the full benefits of html handling in the template and get rid of another use of `create_echo_form`
(naming of the vars are up to you, I just tried to choose one that made sense)

templates/Default/admin/Default/1.6/universe_create_planets.php
```diff
@@ -0,0 +1,23 @@
...
+<table class="standard">
```
Page-: Can you indent the table inside the `<form>` please

admin/Default/1.6/universe_create_save_processing.php
```diff
@@ -167,39 +167,27 @@
...
-//		$updatePlan[$sector]['Owner Type'] = 'Player';
+	$allowedTypeIDs = array();
+	$db->query('SELECT planet_type_id FROM planet_type');
+	while ($db->nextRecord()) {
+		$allowedTypeIDs[] = $db->getInt('planet_type_id');
+	}
+
+	foreach ($allowedTypeIDs as $planetTypeID) {
+		$numberOfPlanets = $_POST['type' . $planetTypeID];
+		for ($i=1;$i<=$numberOfPlanets;$i++) {
```
Page-: Could you change this to
```
for ($i = 0; $i < $numberOfPlanets; $i++) {
```
please.  As `$i` isn't used it's weird to see it start from 1

hemberger: This line was there before I made any changes (line 183), and `$i` wasn't used in the previous version either. I can change it if you want, but the rest of this file does the exact same thing (which would make this line the odd one out). But if it were my choice, I'd rather leave it as is (to address in a later cleanup of this file) to avoid entangling functionality changes with code cleanup. Let me know!

Page-: Yeah, I realised it wasn't a change by you - I didn't realise it's used throughout the rest of this file though, we can't leave it as is so it's at least consistent within the file

PR #134, opened 2017-09-05
Categorize list of locations in galaxy creation

hemberger:
When adding locations to a galaxy, the list of locations is very long, and so it is onerous to scroll through whether you're looking for a specific location or a type of location. By adding navigable categories to the list, we make the process of galaxy creation easier. 

We also convert the page to a template.

image

### DIFF
--- a/admin/Default/1.6/universe_create_locations.php
+++ b/admin/Default/1.6/universe_create_locations.php
@@ -22,51 +22,73 @@ $template->assignByRef('TotalLocs', $totalLocs);
 $galaxy =& SmrGalaxy::getGalaxy($var['game_id'],$var['gal_on']);
 $template->assignByRef('Galaxy', $galaxy);
 
+// Though we expect a location to be only in one category, it is possible to
+// edit a location in the Admin Tools so that it is in two or more categories.
+// For simplicity here, it will only show up in the first category it matches,
+// but it will identify all other categories that it is in.
+// If multi-category locations becomes common, this code should be modified.
+class Categories {
+	public $locTypes = array();
+	private $locAdded = array();  // list of locs added to a category
+	public function addLoc($locID, $category) {
+		if (!$this->added($locID)) {
+			$this->locTypes[$category][] = $locID;
+			$this->locAdded[] = $locID;
+			return '';
+		} else {
+			return "<b>Also in $category</b><br />";
+		}
+	}
+	public function added($locID) {
+		return in_array($locID, $this->locAdded);
+	}
+}
+
 // Set any extra information to be displayed with each location
 $locText = array();
-$locTypes = array();
+$categories = new Categories();
 foreach ($locations as &$location) {
 	$extra = '<span class="small"><br />';
 	if ($location->isWeaponSold()) {
-		$locTypes['Weapons'][] = $location->getTypeID();
+		$extra .= $categories->addLoc($location->getTypeID(), 'Weapons');
 		$weaponsSold =& $location->getWeaponsSold();
 		foreach($weaponsSold as &$weapon) {
 			$extra .= $weapon->getName() . '&nbsp;&nbsp;&nbsp;(' . $weapon->getShieldDamage() . '/' . $weapon->getArmourDamage() . '/' . $weapon->getBaseAccuracy() . ')<br />';
 		} unset($weapon);
 	}
-	elseif ($location->isShipSold()) {
-		$locTypes['Ships'][] = $location->getTypeID();
+	if ($location->isShipSold()) {
+		$extra .= $categories->addLoc($location->getTypeID(), 'Ships');
 		$shipsSold =& $location->getShipsSold();
 		foreach ($shipsSold as &$shipSold) {
 			$extra .= $shipSold['Name'] . '<br />';
 		} unset($shipSold);
 	}
-	elseif ($location->isHardwareSold()) {
-		$locTypes['Hardware'][] = $location->getTypeID();
+	if ($location->isHardwareSold()) {
+		$extra .= $categories->addLoc($location->getTypeID(), 'Hardware');
 		$hardwareSold =& $location->getHardwareSold();
 		foreach ($hardwareSold as &$hardware) {
 			$extra .= $hardware['Name'] . '<br />';
 		} unset($hardware);
 	}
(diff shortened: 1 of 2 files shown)


templates/Default/admin/Default/1.6/universe_create_locations.php
```diff
@@ -0,0 +1,43 @@
...
+
+<?php
+foreach ($LocTypes as $category => $LocIDs) { ?>
+		<a href="javascript:;" onclick="window.location.hash='<?php echo $category; ?>'">Jump to <?php echo $category; ?></a><br /><?php
+} ?>
+
+<br />
+Click a category heading to toggle its display.
+
+<?php echo $Form; ?>
```
Page-: Can you change this to
```php
$template->assign('CreateLocationsFormHREF', SmrSession::getNewHREF($container));
...
<form method="POST" action="<?php echo $CreateLocationsFormHREF; ?>">
```
so we can get the full benefits of html handling in the template and get rid of another use of `create_echo_form`
(naming of the vars are up to you, I just tried to choose one that made sense)

admin/Default/1.6/universe_create_locations.php
```diff
@@ -1,66 +1,80 @@
...
+		$locTypes['Weapons'][] = $location->getTypeID();
 		$weaponsSold =& $location->getWeaponsSold();
 		foreach($weaponsSold as &$weapon) {
 			$extra .= $weapon->getName() . '&nbsp;&nbsp;&nbsp;(' . $weapon->getShieldDamage() . '/' . $weapon->getArmourDamage() . '/' . $weapon->getBaseAccuracy() . ')<br />';
 		} unset($weapon);
-			//$extra .= '<tr><td class="right"><span class="small">' . $WEAPONS[$wep_id]['Weapon Name'] . '</span></td><td class="left"><span class="small">(' . $WEAPONS[$wep_id]['Shield Damage'] . '/' . $WEAPONS[$wep_id]['Armor Damage'] . '/' . $WEAPONS[$wep_id]['Accuracy'] . ')</span></td></tr>';
-		//$extra .= '</table>';
 	}
-	if ($location->isShipSold()) {
+	elseif ($location->isShipSold()) {
```
Page-: This change means that any shop that can sell multiple things will only show one type of thing that it can sell - is it possible to avoid that?

templates/Default/admin/Default/1.6/universe_create_locations.php
```diff
@@ -1,11 +1,16 @@
-Working on Galaxy : <?php echo $Galaxy->getName(); ?> (<?php echo $Galaxy->getGalaxyID() ?>)<br />
+Working on Galaxy : <?php echo $Galaxy->getName(); ?> (<?php echo $Galaxy->getGalaxyID() ?>)<br /><br />
+
+<?php
+foreach ($LocTypes as $category => $LocIDs) { ?>
+		<a href="javascript:;" onclick="window.location.hash='<?php echo $category; ?>'">Jump to <?php echo $category; ?></a><br /><?php
```
Page-: Can you change this to
```html
<a href="'<?php echo $category; ?>">Jump to <?php echo $category; ?></a>
```
please, so that the jumping can be done natively, without needing javascript

hemberger: Unfortunately, linking to anchors using html will only work once. After you click one of these links, you will not be able to click another (see, e.g., https://stackoverflow.com/a/3163635). The javascript does not have this restriction.

If this were a part of the main game, I'd be with you 100% on insisting that there be a non-javascript alternative. But since this is just a convenience item for an admin tool, it seems harmless to me. But if you know how to implement the html workarounds to the one-click limitation, please go for it!

admin/Default/1.6/universe_create_locations.php
hemberger: If a shop sells items from more than one category, then the whole categorization scheme breaks down. Though it is not an enforced rule, the game fortunately follows it.

If for some reason we decide to add a location that violates that rule, then what do we do? Do we display it in only one of the categories (and have people unable to find it if they are looking for it in another category)? Do we display it in multiple categories and then implement some logic to make sure it's not double-counted? It's a tricky situation once this rule is violated...

Given the complexity of the game and the rate of code development, it seems reasonable to code for the game we have rather than the game we _could_ have in the future (but only if it makes the current game much more manageable, which in the case of this list, I bet anyone who has created a game would agree it does).

If you decide to go with this as is, I'd be happy to add a comment in the code explicitly pointing out this restriction. Then if someone decides to violate it in the future, they'll see this and need to think about how to make this list manageable under the new assumptions.

templates/Default/admin/Default/1.6/universe_create_locations.php
Page-: @hemberger that link doesn't seem to suggest any problem with anchor # links, and I've never been aware of any issues (it certainly works fine in chrome on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a )

hemberger: I was referring to:
image
If I use the code you suggested, the link can be clicked once, but after that anchor links will do nothing when clicked. I tested it to confirm, and I'm fairly certain I set it up correctly -- I can set up again on the test server to show you if you'd like.

admin/Default/1.6/universe_create_locations.php
Page-: Well the thing is that there's an admin tool right now that lets admins edit locations, so it is already in the game (albeit whether it's actually used is another matter) and this tools supports seeing it.  I think the easiest solution for the categorisation is just to categorise as the first type of good in the list, and then add the total list later, nicest way I can think to do that is something like:
```php
function addCategory(&$added, &$arr, $typeID) {
    if (!$added) {
        $arr[] = $typeID;
    }
    $added = true;
}

foreach ($locations as &$location) {
	$categoryAdded = false
	if ($location->isWeaponSold()) {
		addCategory(categoryAdded, $locTypes['Weapons'], $location->getTypeID())
		$locTypes['Weapons'][] = $location->getTypeID();
		...

hemberger:
For multi-category locations, I used your idea, but also added a little bit of extra text to identify the additional categories:
image

PR #136, opened 2017-09-10
SectorPlayers.inc: change color of protected allies

hemberger:
Currently, players are displayed in sector with the following colors:

Enemies (newbie turns=NO):  red
Enemies (newbie turns=YES): yellow
Allies (newbie turns=NO):   green
Allies (newbie turns=YES):  green

However, this often causes confusion when an ally is in newbie turns.
It can make it difficult to count the number of allies who are able
to fight or to determine if an ally has forgotten to leave newbie
protection.

We change the colors here to make it immediately visible if an ally
is in newbie turns. Since you can already click "Examine" to inspect
if the ally has newbie turns (it will say "Your target is under newbie
protection!"), this change does not modify what information is
available to players -- only how it is displayed.

Display players with the following new colors:

Enemies (newbie turns=NO):  red
Enemies (newbie turns=YES): yellow
Allies (newbie turns=NO):   green
Allies (newbie turns=YES):  yellow  <-- changed

### DIFF
--- a/templates/Default/engine/Default/includes/SectorPlayers.inc
+++ b/templates/Default/engine/Default/includes/SectorPlayers.inc
@@ -1,14 +1,14 @@
 <?php
 function getPlayerOptionClass(&$player, &$other) {
 	// Returns the CSS relational class of player "other" relative to "player".
-	if (!$player->traderNAPAlliance($other)) {
-		if ($other->canFight()) {
-			return "enemy";
+	if ($other->canFight()) {
+		if ($player->traderNAPAlliance($other)) {
+			return "friendly";
 		} else {
-			return "neutral";
+			return "enemy";
 		}
 	} else {
-		return "friendly";
+		return "neutral";
 	}
 }
 ?>
--- a/templates/Default/engine/Default/includes/SectorPlayers.inc
+++ b/templates/Default/engine/Default/includes/SectorPlayers.inc
@@ -1,3 +1,18 @@
+<?php
+function getPlayerOptionClass(&$player, &$other) {
+	// Returns the CSS relational class of player "other" relative to "player".
+	if (!$player->traderNAPAlliance($other)) {
+		if ($other->canFight()) {
+			return "enemy";
+		} else {
+			return "neutral";
+		}
+	} else {
+		return "friendly";
+	}
+}
+?>
+
 <div id="players_cs" class="ajax"><?php
 	if($PlayersContainer->hasOtherTraders($ThisPlayer)) {
 		$Players =& $PlayersContainer->getOtherTraders($ThisPlayer);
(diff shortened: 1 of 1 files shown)


hemberger:
Original colors:
image

New colors:
image

templates/Default/engine/Default/includes/SectorPlayers.inc
```diff
@@ -1,3 +1,18 @@
+<?php
+function getPlayerOptionClass(&$player, &$other) {
```
Page-: Could you rename this to something like `getDiplomaticStatus` or something and add it to `SmrPlayer` as eg `$player->getDiplomaticStatus($otherPlayer)`, as it's something that would make sense to live there, and gets this logic out of the template

hemberger:
The requested changes have been made. I'm not a huge fan of the name I chose, `SmrPlayer.getPlayerRelationship`, but everything else I could think of seemed like it would be too easy to mistake for a racial council function. I'm open to alternatives.

lib/Default/SmrPlayer.class.inc
```diff
@@ -465,16 +465,17 @@ class SmrPlayer extends AbstractSmrPlayer {
 		$this->hasChanged=true;
 	}
 
-	function getAttackColour() {
```
Page-: Could you readd these or remove them completely please (I know these functions are not used, but if we're gonna have the db fields/class properties then it would be nice to keep the methods available to make use of, or just get rid of it altogether).  For reference these were a 1.5 thing where you could choose the colour of the flashing screen when you get attacked

hemberger: We already have one vote for keeping it (and even re-implementing the ability to customize the color), so unless you object I'll keep that code in there.

hemberger:
One additional thought I had: should we make the same color changes to the local map? It might be confusing to have the two be different.

lib/Default/SmrPlayer.class.inc
Page-: Keeping it works for me

Page-:
I don't think the local map has an idea of neutral does it? If it does then it could make sense but I think I avoided it as the amount of db requests it would have to make to check every player in every surrounding sector could be fairly slow

hemberger:
The local map does in fact check for friendly/enemy/neutral players [here](https://github.com/smrealms/smr/blob/d526cb05c81f00f93267837816f11c16cf1a8a91/templates/Default/engine/Default/includes/SectorMap.inc#L95-L118). So unless I hear otherwise, I'll go ahead and re-use the logic from the current sector in the local map.

hemberger:
The code change for the local map won't affect performance. However, this is actually a little more complicated than I expected. In the piece of code I linked above ([here](https://github.com/smrealms/smr/blob/d526cb05c81f00f93267837816f11c16cf1a8a91/templates/Default/engine/Default/includes/SectorMap.inc#L97-L107)), we can see how interlinked the idea of "friendly" and "allied" are. In the Current Sector screen, what you really want to see is "can this trader attack with me?" (because you can see the alliances -- there is no degeneracy between protected allies and protected enemies). However, that degeneracy _does_ exist on the Local Map. So how do we reconcile this?

What we want:
* Consistent colors between Local Map and Current Sector
* A way to distinguish between protected and unprotected traders in the Current Sector (achieved by the original PR)
* A way to distinguish between allied and enemy _and_ protected and unprotected traders in the Local Map

So how do we do this, given that we only have 3 colors, but 4 states we want to identify (protected/unprotected ally, protected/unprotected enemy)?

Or do we want to argue that the Local Map has reduced information, and you shouldn't be able to identify all 4 states? In that case, I assume we would still want the colors to be consistent with the Current Sector, and that requires a restructuring of the logic in the code I linked. It also raises questions like: Do we want to see the location of all allied traders on the galaxy/local map, or do we want to see the location of all "friendly" (in the new definition, i.e. can attack with you) traders. If the former, then that means you would see both green and yellow icons in any sector (and thus a bit of a change in the code logic).

Thoughts?

Page-:
I think it's fine for the local map to have less granular info because from a roleplaying sense the info you're getting is via a longer range scan that isn't able to get all of the info (but for allies you can get more info via comms/whatever).  Overall I think it's your call though as you know the state of the current game far better than I do

witold-cyrek:
Is this finished? I know the original way was confising but there was a reason for this.  Mainly it was quite a cluster to do it the other way, and there is the fact that allied players show up on Galaxy map, thus having them show up yellow sometimes might be a bit confusing.

hemberger:
I stopped working on this for the exact reason you pointed out -- I couldn't decide how this should behave on the Galaxy Map. I think before we make any further changes, we should decide on a motivation for how scanners behave. (Are the local map scanners the same as the current sector scanners? Are there long-range scanners, and if so should they be a perk of a particular race? Are ally positions broadcast by a different mechanism, or is it still some form of scanning?)

PR #144, opened 2017-11-08
Add docker-compose service for serving documentation

hemberger:
Provides simple infrastructure and configuration for serving the generated documentation.

### DIFF
--- a/docs/docker-compose.yml
+++ b/docs/docker-compose.yml
@@ -1,6 +1,13 @@
 version: '2'
 
 services:
+    smr-docs:
+        image: nginx:alpine
+        volumes:
+            - ./api/html:/usr/share/nginx/html:ro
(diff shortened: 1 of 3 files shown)


docs/docker-compose.yml
```diff
@@ -1,6 +1,13 @@
 version: '2'
 
 services:
+    smr-docs:
+        image: nginx:alpine
+        volumes:
+            - ./api/html:/usr/share/nginx/html:ro
+        ports:
+            - "8081:80"
```
Page-: 8081 is the external ip right?

hemberger: That's right. I hard-coded the external port for now because I don't suspect anyone but me will be using this. But if you want it to be a docker-compose environment variable, I can do that.

PR #147, opened 2017-11-09
Add a BBLink to log for force attack messages

hemberger:
Similar to the addition of combat log links in planet attack messages, we can add a BBLink to force combat logs when forces with scout drones are attacked.

To avoid making this change in both `force_{attack,minefield}_processing.php`, we refactor these two files, and then add an `action` variable to the page, which determines whether the "bump" or "attack" conditions are executed. The refactor is done carefully in a series of 3 commits that first synchronize the two files before removing the near-duplicate.


hemberger:
I forgot to add a screenshot of the new messages:
image
In addition to the combat log link, the following changes were made:
* slight change of wording
* red span added to "attack" for consistency with planet attack messages
* sector ID was changed to a BBLink

engine/Default/forces_attack_processing.php
```diff
@@ -41,35 +58,55 @@
...
 	$attacker->getShip()->decloak();
-	$attacker->setLastSectorID(0);
+	if (!$bump) {
+		$attacker->setLastSectorID(0);
+	}
 } unset($attacker);
 
+// If mines are bumped, the forces shoot first. Otherwise player shoots first.
+if ($bump) {
+	$results['Forces'] =& $forces->shootPlayers($attackers,true);
```
Page-: The second param of both of these `$forces->shootPlayers` can be `$bump`, as that's what it's intended to denote anyway

hemberger: Done!

PR #149, opened 2017-11-10
Security layer for combat log BBCode links

hemberger:
Now that we have a BB-tag that links to sensitive information (in this
case combat logs), since player messages are bbify'd, you could send
a message to another player with "[combatlog=X]", where X is some valid
combat log ID, and the recipient would then have a link to that combat
log -- even if they would otherwise be unable to see it in the combat
log viewer.

This exploit is unlikely to be discovered (and even harder to abuse),
but since it is theoretically possible, it needs to be prevented.

To do this, we create a security layer between the BBLink and the
displayed combat log page. This layer verifies that the player should
be allowed to access it. Currently, restrictions include:

* Log is from the current game
* Attacker/Defender is the player or in the player's alliance

### DIFF
--- a/engine/Default/combat_log_viewer_verify.php
+++ b/engine/Default/combat_log_viewer_verify.php
@@ -1,0 +1,25 @@
+<?php
+// Verify that the player is permitted to view the requested combat log
+// Qualifications:
+//  * Log must be from the current game
+//  * Attacker or defender is the player OR in the player's alliance
+
+$query = 'SELECT log_id FROM combat_logs WHERE log_id=' . $db->escapeNumber($var['log_id']) . ' AND game_id=' . $db->escapeNumber($player->getGameID()) . ' AND ';
+if ($player->hasAlliance()) {
+	$query .= '(attacker_alliance_id=' . $db->escapeNumber($player->getAllianceID()) . ' OR defender_alliance_id=' . $db->escapeNumber($player->getAllianceID()) . ')';
+} else {
+	$query .= '(attacker_id=' . $db->escapeNumber($player->getAccountID()) . ' OR defender_id=' . $db->escapeNumber($player->getAccountID()) . ')';
+}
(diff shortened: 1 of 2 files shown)


engine/Default/combat_log_viewer_verify.php
```diff
@@ -0,0 +1,25 @@
+<?php
+// Verify that the player is permitted to view the requested combat log
+// Qualifications:
+//  * Log must be from the current game
+//  * Attacker or defender is the player OR in the player's alliance
+
+$query = 'SELECT log_id FROM combat_logs WHERE log_id=' . $db->escapeNumber($var['log_id']) . ' AND game_id=' . $db->escapeNumber($player->getGameID()) . ' AND ';
+if ($player->hasAlliance()) {
+	$query .= '(attacker_alliance_id=' . $db->escapeNumber($player->getAllianceID()) . ' OR defender_alliance_id=' . $db->escapeNumber($player->getAllianceID()) . ')';
+} else {
```
Page-: This `else` would mean that a player isn't allowed to see their logs that happened whilst they were in another alliance right? Because the alliance id in the logs is the one at the time they attack occurred and I'm fairly sure it doesn't get updated - at the same time the combat logs list seems to do the same thing so maybe it's ok? or it's just bugged there too

hemberger: That's right. The logic is the same as in `combat_log_list.php`, where the query is based solely on the alliance ID if the player is in an alliance (except for the "Personal" combat log).

While the player getting this message is guaranteed to be the "defender" for force attack messages, it is not so for planet attack messages. So I think in this case we need to keep it consistent with `combat_log_list.php`.

I can imagine a more sophisticated query, e.g. (attacker_alliance OR defender_alliance OR attacker OR defender), so that you can still view combat logs you were the attack or defender even if you're no longer in the same alliance, but that potential change seems independent of plugging this current exploit.

PR #150, opened 2017-11-10
Improve procedural message cosmetics

hemberger:
We add two new non-player "accounts" (rather, an account ID that is associated with a non-player sender for procedurally generated messages):

* Scout Drone -- will be the sender displayed on force pings
* Federal Clerk -- will be the sender displayed on force/planet/player kills/deaths (note that port deaths are still sent by "Port Defenses", so it's unclear if that should change)

A few other related changes:

* Planet messages (rather than player messages) are sent to players who are kicked from a planet.
* We improve the information provided in these messages so that it's more consistent and has more BBLinks to players/sectors.
* Move where the report button is in the message headers so that it lines up correctly for all message types.
* Force change messages are now in first person.

This is a rework of #146, except only cosmetic changes.


hemberger:
New position of "notify admin" button:
image

New "kicked from planet" message (note that it's a Planet Message now):
image

New kill/death messages:
image
image

New force ping/attack messages:
image

Force change messages:
image

hemberger:
I realized that the "Scout Drone" changes mess up grouped messages, so I may break this PR into smaller chunks.

engine/Default/forces_drop_processing.php
```diff
@@ -167,7 +167,7 @@
 		$scout_drones_message .= 's';
 
 	// now compile it together
-	$message = $player->getPlayerName().' has ' . $mines_message;
+	$message = 'I have ' . $mines_message;
```
Page-: This one's a little odd, whilst it makes sense from a grammatical point of view to use "I have" as the message is sent from the player but from a gameplay/information sense I think it's nice to have the player name repeated as part of the message rather than having to look outside of the main message for that info

hemberger: Well, it follows the pattern of the alliance join/leave messages, which are in first person. And if we're making a point that it's a message sent by your ally (especially if scout messages end up being sent from a "Scout Drone" rather than the pinging player), then third person looks odd. When you receive an e-mail from someone, it's hopefully not in third person. :D 

But I'll take your preference into consideration as I figure out what to do about the scout message grouping issue. Depending on the solution I come up with (if any), this may be a moot point. I'll ask you about it again once this PR feels more complete.

Page-: Yeah, it's not a huge deal to change it, I just noticed it as I used to use that whilst scanning through messages, but I don't know whether that's a thing that helps others as well or not

PR #151, opened 2017-11-12
Fix posting GP article to Breaking News

hemberger:
* Avoid re-inserting article into news on ajax updates
* Replace link with a "success" message once news is added to breaking news
* Tidy up some whitespace

### DIFF
--- a/engine/Default/galactic_post_view_article.php
+++ b/engine/Default/galactic_post_view_article.php
@@ -8,6 +8,7 @@ if (isset($var['news'])) {
 		'VALUES(' . $db->escapeNumber($player->getGameID()) . ', ' . $db->escapeNumber(TIME) . ', ' . $db->escape_string($var['news'], false) . ', \'BREAKING\')');
 	// avoid multiple insertion on ajax updates
 	SmrSession::updateVar('news', null);
+	SmrSession::updateVar('added_to_breaking_news', true);
 }
 $db->query('SELECT * FROM galactic_post_article WHERE game_id = ' . $db->escapeNumber($player->getGameID()));
 if ($db->getNumRows()) {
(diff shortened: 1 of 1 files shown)


hemberger:
The new/adjusted content in this PR:
image

engine/Default/galactic_post_view_article.php
```diff
@@ -66,13 +69,18 @@
...
-	$container = array();
-	$container['url'] = 'skeleton.php';
-	$container['body'] = 'galactic_post_view_article.php';
-	$container['news'] = $message;
-	transfer('id');
-	$PHP_OUTPUT.=('<small><br />note: breaking news is in the news section.<br /></small>');
-	$PHP_OUTPUT.=create_link($container, 'Add to Breaking News');
+
+	// breaking news options
+	$PHP_OUTPUT.=('<br />');
```
Page-: Btw the brackets aren't a style thing or anything, it's just it used to be direct `print()` calls so when I did a mass replace to `$PHP_OUTPUT` the brackets were left behind (not that they're necessary for `print` either, but they were there) - I don't care whether they're here or not, but it's just a side effect of how I made transition to the new template system be able to be done gradually

hemberger: Understood! I try to keep the formatting consistent in a given file, even if it's not ideal. But I imagine these superfluous parentheses will naturally disappear when this page is converted into a template.

Page-: Yeah, I just wanted to clarify that it's an artifact of the conversion process rather than being an intended thing, and the files that don't have it are likely because it was pretty wildly inconsistent whether parens were used or not before, probably depending on which Mike coded it

PR #155, opened 2017-11-14
Improve "Request A Feature" pages

hemberger:
Public improvements:
* Requests/comments no longer anonymous by default (checkbox is still there)
* Votes are color-coded

Moderator improvements:
* Option to set feature request status while viewing comments
* Disentangle "Vote" and "Set Status" buttons, so you can only do one or the other at a time
* Use a dummy "Select Status" default option for the status dropdown

Code improvements:
* Fix a couple misleading variable names

### DIFF
--- a/templates/Default/engine/Default/feature_request.php
+++ b/templates/Default/engine/Default/feature_request.php
@@ -55,8 +55,14 @@ if(isset($FeatureRequests)) { ?>
 			} unset($FeatureRequest); ?>
 		</table>
 		<div align="right"><?php
-			if($FeatureModerator) {
-				?>&nbsp;<select name="status"><option value="Implemented">Implemented</option><option value="Rejected">Rejected</option><option value="Opened">Open</option><option value="Deleted">Delete</option></select>&nbsp;<input type="submit" name="action" value="Set Status"><?php
+			if($FeatureModerator) { ?>&nbsp;
+				<select name="status">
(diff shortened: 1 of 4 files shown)


templates/Default/engine/Default/feature_request.php
```diff
@@ -74,11 +80,11 @@
 				<td align="center"><textarea spellcheck="true" name="feature" id="InputFields" maxlength="500"></textarea></td>
 			</tr>
 			<tr>
-				<td align="center">Anonymous: <input name="anon" id="InputFields" type="checkbox" checked="checked"/></td>
+				<td align="center">Anonymous: <input name="anon" id="InputFields" type="checkbox" /></td>
```
Page-: I'd rather keep this as default anonymous, the idea is to allow people to safely request features and changes without having to worry about the idea being rejected due to the person who requested it rather than due to the idea's merits or lack thereof , and without having to worry about being harassed by other players in the game about it (which can and does happen, unless the playerbase has massively changed in the past few years).  And the reason it needs to be the default rather than opt-in is because if you forgot to tick it you can't then make it anonymous, however you can then make a comment claiming it as yours

hemberger: Fair enough! I'll revert that commit now.

hemberger:
Removed the change to the anonymous checkbox. It remains checked by default.

PR #158, opened 2017-11-14
Fix an assortment of PHP warnings and related minor bugs

hemberger:
* Fix an assortment of PHP warnings that were dominating the nginx logs
* Fix identification of non-player alliance message thread authors (e.g. Bank Reporter)

### DIFF
--- a/engine/Default/bank_report.php
+++ b/engine/Default/bank_report.php
@@ -16,23 +16,15 @@ $db->query('SELECT * FROM alliance_bank_transactions WHERE alliance_id = ' . $db
 if (!$db->getNumRows()) {
 	create_error('Your alliance has no recorded transactions.');
 }
+$trans = array();
 while ($db->nextRecord()) {
-	if ($db->getField('transaction') == 'Payment') {
-		if (!$db->getField('exempt')) {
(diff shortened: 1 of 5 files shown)


engine/Default/bank_report.php
```diff
@@ -16,23 +16,15 @@
...
-	}
-	else {
-		if (!$db->getField('exempt')) {
-			$trans[$db->getInt('payee_id')][DEPOSIT] += $db->getInt('amount');
-		}
-		else {
-			$trans[0][DEPOSIT] += $db->getInt('amount');
-		}
+	$transType = ($db->getField('transaction') == 'Payment') ? WITHDRAW : DEPOSIT;
+	$payeeId = ($db->getField('exempt')) ? 0 : $db->getInt('payee_id');
```
Page-: Should this be `getBoolean` instead of `getField`?

hemberger: For some reason the `exempt` field is a `tinyint(1)` instead of a bool. But if `getBoolean` would correctly convert it (even if the value is not 1/0 or true/false), then it would certainly make the code clearer.

Page-: Oh in that case it won't :(

PR #166, opened 2017-11-18
current_players.php: add link to send global message

hemberger:
This feature was requested in 2011! They pointed out that
it may allow new players to more easily get help.


hemberger:
image

PR #167, opened 2017-11-19
Add Discord ID fields to the database

hemberger:
Store a Discord Channel in SmrAlliance

For now, we restrict to a 1-to-1 mapping between accounts and
Discord Channel ID's for any given game. We allow duplicates
across games since alliances may use the same channel for
multiple games.

* SmrAlliance: add `{get,set}DiscordChannel` methods
* MySQL database: add `discord_channel` field to `alliance` table
* Alliance Stats: add field to change/delete Discord Channel ID

----------------------

  Store a Discord User ID in SmrAccount

For now, we restrict to a 1-to-1 mapping between accounts and
Discord ID's. We will handle the case of multi support later.

* SmrAccount: add `{get,set}DiscordId` methods
* MySQL database: add `discord_id` field to `account` table
* Preferences: add field to change/delete Discord ID

-------------------------

These changes require a flyway migration.

We add the Discord User and Channel ID's to the database now in preparation
for forthcoming Discord integration. Having the database updated now will make
it easier to test the new infrastructure.

The Discord integration is completed and tested locally, but needs some careful
testing with the software stack on the server since it is not using Docker yet.

### DIFF
--- a/db/patches/V1_6_61_06__discord_id.sql
+++ b/db/patches/V1_6_61_06__discord_id.sql
@@ -1,0 +1,2 @@
+-- Add `discord_id` field to `account` table
+ALTER TABLE `account` ADD COLUMN `discord_id` VARCHAR(32) DEFAULT NULL AFTER `hof_name`;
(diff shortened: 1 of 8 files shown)


hemberger:
In User Preferences:
image

In Change Alliance Stats:
image

PR #170, opened 2017-11-21
SmrPlayer: change LevelName behavior

hemberger:
Previously, `SmrPlayer::getLevelName` would return only your
presidential status if you were a racial president, making it
difficult to enjoy your rise through the experience ranks.

Displaying both rank and council title was an often suggested
feature. So instead of replacing the experience-based rank with
the title, we use a small 16x14 crown image (based on the crown
in the council message image).

There was also some outdated placeholder replacements in
`getLevelName`. Since these are no longer (or were never?) used,
I removed them.


hemberger:
image
And when you hover over the image, it tells you the presidential title:
screenshot_2017-11-20_17-20-42

PR #171, opened 2017-11-21
Revamp the port attack page

hemberger:
* Replace textual scan results ("Your scanners detect...") with a
  table that looks like the sector scan results. It gives the page
  a little more substance and also makes it more consistent with
  how other pages look.

* Add a short bit of roleplaying text to help newer players realize
  that raiding a port might not be a good idea. A link to the wiki
  port combat page is provided. This was tip for newbies was
  frequently requested by players.

* Add number of turrets to the port scan.

### DIFF
--- a/lib/Default/SmrPort.class.inc
+++ b/lib/Default/SmrPort.class.inc
@@ -43,9 +43,6 @@ class SmrPort {
 	protected $experience;
 	protected $goods;
 	
-	protected $turrets = 20;
-	protected $turretDamage = 250;
-	
 	protected $cachedVersion = false;
 	protected $cachedTime = TIME;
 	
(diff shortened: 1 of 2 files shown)


hemberger:
Without a scanner:

image

With a scanner:
image

lib/Default/SmrPort.class.inc
```diff
@@ -813,11 +810,16 @@ class SmrPort {
...
-	
+
+	public function getNumWeapons() {
+		return self::NUMBER_OF_PORT_WEAPONS;
+	}
+
 	public function &getWeapons() {
 		$weapons = array();
-		for($i=0;$i<self::NUMBER_OF_PORT_WEAPONS;++$i)
+		for ($i=0; $i<$this->getNumWeapons(); ++$i) {
```
Page-: This function will be called once for every iteration of the loop, it's not a big deal because it's a cheap function but something to be aware in other situations/if the function becomes more expensive

PR #172, opened 2017-11-21
Add Discordbot

hemberger:
Implemented using AnubisBot, which is a light wrapper around
the DiscordPHP library.

Unfortunately, DiscordPHP is a dead project, but since we only
use basic messaging functionality, I don't think we should encounter
any insurmountable problems going forward.

### DIFF
--- a/tools/discord/Dockerfile
+++ b/tools/discord/Dockerfile
@@ -1,0 +1,31 @@
+FROM php:7-cli
+RUN apt-get update \
+	&& apt-get install -y git \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& docker-php-ext-install mysqli
+
+# Install composer
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
+
+# Install anubisbot
+RUN composer create-project pazuzu156/anubisbot --prefer-dist
+
+# Install SMR-related dependencies
+WORKDIR /smr
+COPY tools/discord/composer.json .
+RUN composer install --no-interaction
+
+# Get the SMR source code
+COPY . .
+
+# Set up anubisbot commands
+WORKDIR /anubisbot
+RUN php ./cli make:command Turns
+RUN php ./cli make:command Money
+RUN php ./cli make:command Game
+
+COPY tools/discord/cli .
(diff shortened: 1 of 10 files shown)


hemberger:
This version of Discordbot will be superseded by one that uses DiscordPHP directly, rather than AnubisBot.

hemberger:
For future reference, this code had several major issues:
1. The debugging output could not be turned off (see https://github.com/pazuzu156/AnubisBot/issues/26).
2. Whether due to an issue with PHP 7 / mysqli or due to AnubisBot itself, errors within the registered commands would silently abort. The only way to circumvent this was to wrap every command in a try-catch (see https://github.com/pazuzu156/AnubisBot/issues/25).
3. Whether due to an issue with mysqli or due to AnubisBot itself, the database connection would timeout over the course of 1 day, leaving the bot running but unable to respond to commands.
4. It could not be used with PHP 5.6.

pazuzu156:
Necro-bumping, but meh. I feel I should point out AnubisBot (the PHP port) is as dead as DiscordPHP. I've already ported over to C# and basically flagged the PHP code as un-maintained and not suitable for use. Unfortunately, it got to the point I was beginning to make more into a library to circumvent some of DiscordPHP's issues, and ultimately became too much. DiscordPHP was also removed from Discord's resources page, so use of the library is not advised unless you can find a version that uses v7 of the Discord API

hemberger:
Hi, and thanks for your comments! Yeah, I understand that DiscordPHP is dead, and that some of the issues I mentioned above were issues within DiscordPHP itself, but since we have PHP library calls we need to make within the bot events, it makes the most sense to continue to use DiscordPHP until it no longer works.

If you have any suggestions for the best way to switch to a better-supported Discord API when you have PHP libraries that must remain in PHP, I'd be very appreciate of any advice you can give!

pazuzu156:
It seems there's one that is still being updated, so it likely supports v7 of the Discord API. I don't know how close it'll resemble DiscordPHP, but it's worth a look .Especially when Discord decides to shut down v6 of the API.

https://github.com/CharlotteDunois/Yasmin

hemberger:
I'll check it out, thanks!

PR #173, opened 2017-11-26
Add page to view past Galactic Post editions

hemberger:
* Reconfigure how online papers are stored in the database. We move
  the `online_since` column of the `galactic_post_online` table to
  the `galactic_post_paper` table. If `online_since` is NULL, then
  the paper is not published. This allows us to track publication
  history, which is needed for listing past editions.

  Previously, the publication history was overwritten. Since this
  is no longer the case, we need to select the "online" paper as the
  one with the most recent `online_since` value.

* Refactor galactic_post_read.php to take parameters `game_id` and
  `paper_id` so that it can be used for current and past editions.

* New galactic_post_current.php calls galactic_post_read.php with
  the player's Game ID and most recently published paper ID.

* New Galactic_post_past.php calls galactic_post_read.php with
  the selected paper's game and paper IDs.

We allow viewing past game Galactic Post editions because it
provides a lot of content for players to read, and it can be
very nostalgic to look back. Limiting to the current game would
make more sense, but is much less useful. It also gives players
more incentive to write articles, since they will be persistent.

To regain publication history from games prior to this change, we
may want to manually update `online_since` for papers that were
likely published (since `galactic_post_online` only kept track of
the most recently published paper of each game).


hemberger:
Added some cosmetic (and visual robustness) changes to the editor options page:

image

engine/Default/galactic_post_delete.php
```diff
@@ -5,19 +5,13 @@
...
-		$db->query('SELECT * FROM galactic_post_paper_content WHERE game_id = ' . $db->escapeNumber($player->getGameID()) . ' AND paper_id = '.$db->escapeNumber($var['id']));
-		while($db->nextRecord()) {
-			//delete this article that is part of this paper
-			$db2->query('DELETE FROM galactic_post_article WHERE article_id = ' . $db->escapeNumber($db->getInt('article_id')) . ' AND game_id = ' . $db->escapeNumber($player->getGameID()));
-		}
+	//find out which articles need to be deleted as well
+	$db->query('SELECT * FROM galactic_post_paper_content WHERE game_id = ' . $db->escapeNumber($player->getGameID()) . ' AND paper_id = '.$db->escapeNumber($var['id']));
+	while($db->nextRecord()) {
+		//delete this article that is part of this paper
+		$db2->query('DELETE FROM galactic_post_article WHERE article_id = ' . $db->escapeNumber($db->getInt('article_id')) . ' AND game_id = ' . $db->escapeNumber($player->getGameID()));
```
Page-: You can make this into a single query with:
```sql
DELETE FROM galactic_post_article gpa
WHERE EXISTS (
	SELECT 1
	FROM galactic_post_paper_content gpc
	WHERE gpa.game_id = gpc.game_id
	AND gpa.article_id = gpc.article_id
	AND gpc.game_id = :game_id
	AND gpc.paper_id = :paper_id
)

File: engine/Default/galactic_post_current.php [hunk 2]
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$db->query('SELECT * FROM galactic_post_paper WHERE online_since IS NOT NULL AND game_id = ' . $db->escapeNumber($player->getGameID()) . ' ORDER BY online_since DESC');
```
Page-: It'd be worth adding a `LIMIT 1` so that we only fetch the first row rather than all of the matching ones

engine/Default/galactic_post_past.php
```diff
@@ -0,0 +1,40 @@
...
+if (isset($_POST['game_id'])) {
+	SmrSession::updateVar('game_id', $_POST['game_id']);
+} else if (!isset($var['game_id'])) {
+	SmrSession::updateVar('game_id', $player->getGameID());
+}
+$template->assign('SelectedGame', $var['game_id']);
+
+// Get the list of games with published papers
+// Add the current game to this list no matter what
+$db->query('SELECT game_name, game_id FROM game WHERE game_id IN (SELECT DISTINCT(game_id) FROM galactic_post_paper WHERE online_since IS NOT NULL) OR game_id=' . $db->escapeNumber($player->getGameID()) . ' ORDER BY game_id DESC');
```
Page-: The brackets after `DISTINCT` are a little confusing, it makes it seem like `DISTINCT` is a function acting only on `game_id` when in reality it's a modifier to the `SELECT` (like `LEFT JOIN` is a modifier to `JOIN`)

engine/Default/galactic_post_past.php
```diff
@@ -0,0 +1,40 @@
...
+$publishedGames = array();
+while ($db->nextRecord()) {
+	$publishedGames[] = array('game_name' => $db->getField('game_name'),
```
Page-: It might be cool if we could do something like
```
$publishedGames = $db->map(function ($db) {
	return array('game_name' => $db->getField('game_name'),
		'game_id' => $db->getInt('game_id'));
});
```
to handle this pattern, although it's not a huge difference in terms of verbosity

Page-: It would be nicer to just use `'name'` and `'id'` here, because the game part will be referenced in the var name, eg `$game['id']` and `$game['name']`

engine/Default/galactic_post_delete.php
hemberger: I actually only changed whitespace on these lines, so I'd prefer not to change this query if possible. But I will mark this down as a future change to make, if that's okay.

engine/Default/galactic_post_past.php
hemberger: Done!

engine/Default/galactic_post_current.php
hemberger: Done!

engine/Default/galactic_post_past.php
hemberger: I agree that if you're looking at the `engine` file, your suggested var names are an improvement because they're less redundant. However, in the `template` file you can't immediately check the query to see what `name` and `id` refer to. You can guess from the name of the variable, but you could also wonder if it is referring to a paper's name and ID (since this code intermingles game and paper queries). I think that hint of doubt warrants the more verbose names.

That said, if you strongly encourage me to change it, I will!

PR #174, opened 2017-11-26
Move path defines to config.php

hemberger:
There is no benefit to specifying the repository path defines
in config.specific.php. These paths cannot be changed, so we
hard-code them into config.php instead of relying on the user
to correctly define them.

Furthermore, having to manually specify `ROOT` can only ever
cause problems. The correct value for `ROOT` can be determined
programmatically using `__DIR__`.

### DIFF
--- a/htdocs/config.php
+++ b/htdocs/config.php
@@ -1,11 +1,22 @@
 <?php
 @date_default_timezone_set(@date_default_timezone_get());
-require_once(dirname(__FILE__) . '/config.specific.php');
 
 if (ENABLE_DEBUG) {
 	// Warn about everything when in debug mode
 	error_reporting(E_ERROR | E_WARNING | E_PARSE | E_NOTICE);
 }
+
+// Repository paths
+define('ROOT', __DIR__ . '/../');
+define('LIB', ROOT.'lib/');
+define('ENGINE', ROOT.'engine/');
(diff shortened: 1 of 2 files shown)


htdocs/config.php
```diff
@@ -1,12 +1,23 @@
...
+define('ROOT', __DIR__ . '/../');
+define('LIB', ROOT.'lib/');
+define('ENGINE', ROOT.'engine/');
+define('WWW', ROOT.'htdocs/');
+define('UPLOAD', WWW.'upload/');
+define('ADMIN', ROOT.'admin/');
+define('TOOLS', ROOT.'tools/');
+
+// Define server-specific constants
+require_once(WWW . '/config.specific.php');
```
Page-: It'd be worth moving the config out of the web dir now, to reduce the chance of leaking the config file contents in case some setting or other is screwed up and the file gets served as text

hemberger: Sounds good to me! Is that coupled to this PR or independent?

Page-: It can be separate

hemberger:
I was trying to find the best place to put it, and nothing seemed natural besides ROOT (which probably isn't a great place to put it either for the sake of avoiding clutter). Would you perhaps want me to move _all_ config files to a `config` subdirectory?

PR #175, opened 2017-11-27
Simplify login creation

hemberger:
* Be less verbose about multiple accounts. It was an intimidating
  paragraph (with some typos too).
* Link to multi rules on wiki instead of the webboard. The previous
  link required a webboard login.
* Since address is optional, simply remove it altogether.
* Make it clearer what the "User Agreement" is.

### DIFF
--- a/htdocs/login_create.php
+++ b/htdocs/login_create.php
@@ -30,11 +30,8 @@ require_once('config.inc');
 
 				<h1>Create Login</h1>
 				<p align='justify'>
-					Creating multiple logins will not be tolerated. If we discover someone playing with several
-					logins, then all of them will be deleted. We have implemented architecture elements to help
-					us detect this. There will be no toleration for multis! If an player is caught using a multi
-					that player's accounts WILL be deleted without ANY warning.<br />
-					<a href='http://smrcnn.smrealms.de/viewtopic.php?t=382' style='font-weight:bold;'>Click HERE</a> for more information.
+					Creating multiple logins is not allowed.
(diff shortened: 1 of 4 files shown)


hemberger:
image
I don't have the recaptcha enabled for my local machine, but it will show up under Last Name.

htdocs/login_create.php
```diff
@@ -30,11 +30,8 @@
 
 				<h1>Create Login</h1>
 				<p align='justify'>
-					Creating multiple logins will not be tolerated. If we discover someone playing with several
-					logins, then all of them will be deleted. We have implemented architecture elements to help
-					us detect this. There will be no toleration for multis! If an player is caught using a multi
-					that player's accounts WILL be deleted without ANY warning.<br />
-					<a href='http://smrcnn.smrealms.de/viewtopic.php?t=382' style='font-weight:bold;'>Click HERE</a> for more information.
+					Creating multiple logins is not allowed.
```
Page-: It's probably worth specifying that it's a single extra account, the current wording implies you can create as many as you want

hemberger: Which wording? On the creation page, the wording is "Creating multiple logins is **not** allowed", since I didn't want to complicate the account creation for newbies, leaving it to the wiki to define the exception. But if you'd prefer an explicit statement about the current rules on that page, I can modify what I currently have there.

Page-: Yeah, most people won't click the link to see how many are actually allowed and assume from the new wording they can just create as many extra as they want and end up getting them all banned

PR #181, opened 2017-11-29
Switch back to upstream list.js

hemberger:
* Move from Page's modified list.js 1.0.0 to the upstream 1.5.0,
  since his PR was (more or less) merged into version 1.1.0.
* naturalSort is exposed from list.utils instead of list.sort.
* sortFunction takes a 3rd option `options` instead of using `this`.


hemberger:
Hang on... somehow I missed a few uses of the removed script.

hemberger:
Okay, should be good to go now!

PR #189, opened 2017-12-02
course_plot_result.php: forward to current sector

hemberger:
If the start sector is not the current sector or if you are on
a planet, then the existing "Plot A Course" results page is
displayed.

Otherwise, you can immediately navigate the plotted course, so go
directly to the current sector page (which includes the plotted
course summary and action buttons) instead of displaying the
intermediate results page.

Add a "Plotted Course Results" heading to the course results
minipage on the current sector page. This will help identify
the course results to anyone who may never have seen the
intermediate course results page before.

This change helps to streamline the flow of the game.

### DIFF
--- a/engine/Default/course_plot_result.php
+++ b/engine/Default/course_plot_result.php
@@ -31,28 +31,10 @@ if ($sector->isLinked($next_sector)) {
 		$player->setPlottedCourse($path);
 	}
 
-	//$PHP_OUTPUT.=create_echo_form($container);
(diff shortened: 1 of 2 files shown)


hemberger:
Changes to plotted course minipage in current sector (with and without scanner):
image

image

hemberger:
Yep, that's a great idea! I will add the `#`.

hemberger:
Modified this to say "Plotted Course" instead of "Plotted Course Results".

PR #190, opened 2017-12-02
Plot To Goods: buy/sell and alignment

hemberger:
Changes to "Plot To Nearest" for Goods:
* Split "Goods" into buy/sell to make the plot feature more useful. This was requested a number of times.
* Respect alignment restrictions for the goods you can plot to.

This PR depends on #188.


templates/Default/engine/Default/course_plot.php
```diff
@@ -67,12 +67,19 @@
 						?><option value="<?php echo $Location->getTypeID(); ?>"><?php echo $Location->getName(); ?></option><?php
 					} unset($Location);
 				break;
-				case 'Goods':
-					$Goods =& Globals::getGoods();
+				case 'Sell Goods':
+				case 'Buy Goods':
+					$Goods =& Globals::getVisibleGoods($ThisPlayer);
```
Page-: It's not huge, but I'd expect this to be:
```php
$Goods =& $ThisPlayer->getVisibleGoods()
```
instead

lib/Default/Plotter.class.inc
```diff
@@ -17,8 +17,18 @@ class Plotter {
 					return SmrLocation::getLocation($X);
 				}
 				return $X;
-			case 'Goods':
```
Page-: It might be worth leaving the generic option in

hemberger: I was having trouble thinking of instances where you want to find a good, but don't care if the port buys or sells it. Do you have any in mind?

templates/Default/engine/Default/course_plot.php
hemberger: Unfortunately there is no such function. There is one for `SmrPort`, but it only returns those goods that are visible that the port buys/sells.

Page-: I know, what I meant was that instead of adding it to `Globals` and requiring a player as an arg that having it in SmrPlayer directly would make more sense

lib/Default/Plotter.class.inc
Page-: The only thing that comes to mind would be route building, but it would be crazy to try doing that via the in-game plot nearest, so nvm

templates/Default/engine/Default/course_plot.php
hemberger: Oh, I see. I didn't add it to `Globals` though. Would you like me to move the method from `Globals` to `SmrPlayer` as part of this PR?

Page-: https://github.com/smrealms/smr/pull/190/files#diff-d812d23765b6f995b38f070caf20222bR202 would beg to differ :P

hemberger: LOL! Is my memory that bad? Okay, I'll move it to player :)

hemberger:
Okay, this PR should be good to go now.

PR #192, opened 2017-12-03
Fix changelog

hemberger:
changelog.php: fix if version table is empty

If the `version` table is empty, then this page doesn't append to
`$PHP_OUTPUT` and therefore gives a blank page (with a PHP warning).

In this case, we simply add a line saying that an admin must
add an initial version to the database first.

--------------------

changelog_view.php: only create one continue button

If there are multiple versions since last login, then a "Continue"
button will show up for each version, which is unnecessary and
also messes up the formatting.

Move the "Continue" button out of the while loop to fix this.

### DIFF
--- a/admin/Default/changelog.php
+++ b/admin/Default/changelog.php
@@ -9,6 +9,11 @@ $link_set_live = true;
 $db->query('SELECT *
 			FROM version
 			ORDER BY version_id DESC');
+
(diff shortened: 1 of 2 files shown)


hemberger:
Merging this now since it is relevant to a short-term bug, but I'll gladly go back and modify anything you're unhappy with here.

PR #196, opened 2017-12-04
Add .dockerignore

hemberger:
Ignoring files that are not needed in the build has three main
advantages:

1. It reduces the size of the build.
2. It allows the build cache to be used if only the ignored
   files change.
3. Ensures that builds in development environments aren't
   contaminated by uncommitted files that may be present.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+# Files to ignore when building Docker image
+
+# Git data directory
+.git
+
+# Committed files that are not needed in the build
+docker-compose.yml
+README.md
+
+# Configuration files will be bind-mounted
+config
+
+# Products from composer install
+vendor
+composer.lock


.dockerignore
```diff
@@ -0,0 +1,19 @@
+# Files to ignore when building Docker image
+
+# Git data directory
+.git
+
+# Committed files that are not needed in the build
+docker-compose.yml
+README.md
+LICENSE
```
Page-: I think the `LICENSE` file should be included, I think technically it would be breaking the license to distribute a docker image without the license included

.dockerignore
```diff
@@ -0,0 +1,19 @@
...
+
+# Uncommitted files that may by lying around
+*.sql
```
Page-: I'd take this one out, it only covers `.sql` files but it would be just as easy to have a `.txt` lying around (eg `notes.txt`, `todo.txt`) or other filetype, and trying to guard against uncommitted files could just bite us later when they turn out to be files that we wanted to include

hemberger:
Okay, the changes you requested were made.

hemberger:
I should note that the fact that we `COPY . .` means that any changes to any files in the repo (committed or not, except for the ones we now ignore) causes a rebuild, which is kind of annoying in a development environment. If you have a `notes.txt` that you modify, you rebuild. If you create a `backup.sql` before you migrate, you rebuild. While this `.dockerignore` goes a long way to addressing these issues, I think that a better long-term solution is moving the entire smr source (e.g. lib, engine, template, admin, tools) into a subdirectory.

PR #197, opened 2017-12-04
alliance_leave_processing.php: fix disbanding

hemberger:
Previously, the player was removed from the alliance, and then
it used the alliance ID of that player to clean up the disbanded
alliance. But since the player was now in alliance 0, nothing
happened.

This fixes disbanding an alliance so that the appropriate db
entries are deleted.

### DIFF
--- a/engine/Default/alliance_leave_processing.php
+++ b/engine/Default/alliance_leave_processing.php
@@ -2,27 +2,33 @@
 $action = $var['action'];
 if ($action == 'YES') {
 	$alliance =& $player->getAlliance();
-	// will this alliance be empty if we leave? (means one member right now)
+
 	if ($player->isAllianceLeader() && $alliance->getNumMembers() > 1) {
 		create_error('You are the leader! You must hand over leadership first!');
 	}
+
+	// will this alliance be empty if we leave? (means one member right now)
+	if ($alliance->getNumMembers() == 1) {
+		// Retain the alliance, but delete some auxilliary info
+		$db->query('DELETE FROM alliance_bank_transactions
+		            WHERE alliance_id = ' . $db->escapeNumber($player->getAllianceID()) . '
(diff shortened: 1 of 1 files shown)


engine/Default/alliance_leave_processing.php
```diff
@@ -2,27 +2,33 @@
...
-	// will this alliance be empty if we leave? (means one member right now)
+
 	if ($player->isAllianceLeader() && $alliance->getNumMembers() > 1) {
 		create_error('You are the leader! You must hand over leadership first!');
 	}
+
+	// will this alliance be empty if we leave? (means one member right now)
+	if ($alliance->getNumMembers() == 1) {
+		//$db->query('DELETE FROM alliance WHERE alliance_id = ' . $db->escapeNumber($player->getAllianceID()) . ' AND game_id = ' . $db->escapeNumber($player->getGameID()));
+		$db->query('DELETE FROM alliance_bank_transactions
```
Page-: I wonder if this is something we actually want to delete, do admins use this to check multi/suspicious behaviour?

hemberger:
At the very least we need to set the `leader_id = 0`. This is very helpful for being able to detect that an alliance is disbanded without checking to see if there are any players left in it. In particular, I need this for `SmrAlliance::getAllianceByDiscordChannel` to filter out disbanded alliances.

I think it would also be fine to delete the alliance threads and the alliance roles. But if you want to keep the bank transactions that seems like it would be fine (though I'm not sure there are any suspicious behavior tools that use the alliance bank transactions, just anon account).

hemberger:
Alternatively I could just unset the discord channel here, rather than setting `leader_id=0`. There are a few places where `leader_id=0` has a special meaning, but it's obviously not necessary for the game to function correctly.

Page-:
I think just setting leader_id to 0 is fine - I think it's the standard check for whether an alliance is disbanded or not

PR #200, opened 2017-12-05
Add chat sharing

hemberger:
Provides a new database table `account_shares_info` and an interface
to manage this table.

This will be used to allow users in chat to display information
about more than just the account associated with their Discord ID
(or IRC Nick). It is generic, but the primary motivation is to
allow players to display their main and multi.

image

image

### DIFF
--- a/engine/Default/chat_sharing_processing.php
+++ b/engine/Default/chat_sharing_processing.php
@@ -1,0 +1,44 @@
+<?php
+
+function error_on_page($message) {
+	$message = '<span class="bold red">ERROR:</span> ' . $message;
+	forward(create_container('skeleton.php', 'chat_sharing.php', array('message' => $message)));
+}
+
+// Process adding a "share to" account
+if (isset($_POST['add'])) {
+	if (empty($_POST['add_player_id'])) {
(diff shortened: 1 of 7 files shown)


engine/Default/chat_sharing_processing.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+function error_on_page($message) {
```
Page-: I'd prefer to use `errorOnPage` to be a bit more consistent, but I am also resigned to the fact that no matter what php isn't a particularly consistent language :(

hemberger: I thought that we were supposed to use camel/pascal case for classes and methods, and snake case for free functions. Or are `create_container`, `create_error`, etc. relics of an earlier standard?

Page-: Yeah, they're relics, a lot of stuff in smr is from times long gone

hemberger: It'd be nice to use something like https://github.com/PHPCheckstyle/phpcheckstyle, but I fear that the initial push for compliance would be extremely burdensome.

PR #203, opened 2017-12-06
changelog_view.php: move continue button to top

hemberger:
Players that haven't logged in for a long time are confronted
with a (potentially) massive list of changes. When the continue
button is at the bottom of the page, this creates a pretty large
barrier of entry.

The continue button at the top of the page ensures that people
a) will immediately know that they have the option to continue
and b) aren't forced to scroll through a list of changes if
they don't care about them.

### DIFF
--- a/engine/Default/changelog_view.php
+++ b/engine/Default/changelog_view.php
@@ -3,11 +3,18 @@
 $template->assign('PageTopic','Change Log');
 
 if(isset($var['Since'])) {
+	$PHP_OUTPUT.=create_button(create_container('logged_in.php'), 'Continue');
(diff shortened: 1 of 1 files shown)


hemberger:
Excellent point. I'll add that back in now!

hemberger:
This is what it looks like with buttons at both top and bottom. Is this okay, or should I find a way to only include the top button if the list is long?

image

hemberger:
That's a good suggestion. Sticky didn't quite work how I wanted, since it would overlap some text, but how about using `overflow-y: auto`?

It creates a scrollbar when the changelog is long
image

And no scrollbar when it is short
image

PR #206, opened 2017-12-07
Default.css: fix persistent vertical scrollbar

hemberger:
No matter how tall you make your browser window, SMR will always
have a vertical scrollbar because the CSS body height was set to
100% with a 10px padding. If we subtract those 10px from the height,
then we no longer get a scrollbar when the browser window is
sufficiently tall.


hemberger:
Before:
image

After:
image

htdocs/originalCSS/Default.css
```diff
@@ -1,7 +1,7 @@
 @import url('shared.css');
 
 html, body {
-	height: 100%;
+	height: calc(100% - 10px); /* Account for padding to fix persistent vertical scrollbar */
```
Page-: `calc` is a pretty new feature ( https://caniuse.com/#feat=calc ) so it may well break for some people, I think it can be done via including both like
```
	height: 100%;
	height: calc(100% - 10px); /* Account for padding to fix persistent vertical scrollbar */
```
and older browsers will use the first because the second one errors (at least it'd be nice if it was that easy)

hemberger: It's supported in IE 9 (2011), Firefox 16 (2012), and Chrome 26 (2013). I feel bad for anyone using browsers older than that. :(

I'll add the fallback as you suggest, which is documented to work: http://html5please.com/#fallback

PR #209, opened 2017-12-09
Reorganize voting for free turns

hemberger:
Changes to voting sites:

* Remove MPOGD, because it no longer exists.

* Re-add TWD, which seems to be up and running just fine.
  I have updated the info for SMR on this site.

Change to code organization:

* Fix a bug where when you weren't in a game, it would always
  display that free turns were available. Now it gives you the
  correct time when not in a game, but still doesn't give you
  the free turns link (since it wouldn't have a game to attribute
  them to).

* Implement a callback for the TWD voting. Free turns are no
  longer awarded automatically. Instead, TWD is set up to callback
  to `vote_link_callback.php` when the player actually votes, and
  turns are awarded there.

  The callback is implemented by passing the game and account ID
  to the vote link in the query string, and TWD will return this
  to us in a POST. Because TWD's vote counting rules are somewhat
  arbitrary, we also pass "alwaysreward" so that free turns are
  consistently awarded for voting.

This patch requires database migration.

-----------------------

Depends on PR #198 to avoid a rebase conflict.

### DIFF
--- a/db/patches/V1_6_61_10__vote_link_callback.sql
+++ b/db/patches/V1_6_61_10__vote_link_callback.sql
@@ -1,0 +1,2 @@
+-- Add turns_claimed column to vote_links table
+ALTER TABLE `vote_links` ADD COLUMN `turns_claimed` enum('TRUE','FALSE') NOT NULL;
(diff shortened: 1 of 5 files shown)


hemberger:
The PR dependency has been merged into master.

PR #210, opened 2017-12-10
Add functionality to manage and respond to ops in-game

hemberger:
* Allow alliance leaders to manage ops in-game in additon to chat.
* When an op is scheduled, a new block of text will be displayed on the alliance MOD page with an option to respond to the op.


hemberger:
Op response on MOD page (appears above other content on page):
image

New option for leader in the "Options" page:
image

Set op page (unscheduled):
image

Set op page (scheduled):
image

PR #214, opened 2017-12-11
Improve bounties display

hemberger:
* Convert `trader_bounties.php` to a template
* Refactor claimable bounties into the `SmrPlayer::getClaimableBounties` method
* Reorganize bounties display in the "Trader Status" and "View Bounties" pages

Trader Status page now displays (less detailed) information about both bounties on your head _and_ bounties you can claim.
image

View Bounties page now displays fully detailed information about both bounties you can claim _and_ bounties on your head.
image

### DIFF
--- a/lib/Default/SmrPlayer.class.inc
+++ b/lib/Default/SmrPlayer.class.inc
@@ -1543,6 +1543,22 @@ class SmrPlayer extends AbstractSmrPlayer {
 		}
 	}
 
+	// Get bounties that can be claimed by this player
+	// Type must be 'HQ' or 'UG'
+	public function getClaimableBounties($type) {
+		$bounties = array();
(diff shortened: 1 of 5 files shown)


lib/Default/SmrPlayer.class.inc
```diff
@@ -1543,6 +1543,22 @@ class SmrPlayer extends AbstractSmrPlayer {
 		}
 	}
 
+	// Get bounties that can be claimed by this player
+	// Type must be 'HQ' or 'UG'
+	public function getClaimableBounties($type) {
+		$bounties = array();
```
Page-: This can be changed to:
```php
return array_filter($this->getBounties(), function($bounty) use  ($type) {
	return $bounty['Type'] === $type;
});

hemberger: Actually, the reason this function is needed as-is is because the many existing bounties methods `getBounties`, etc. based on `getBountiesData`) use `account_id=$this->getAccountID()`, whereas `getClaimableBounties` needs `claimer_id=$this->getAccountID()`.

Perhaps the method name needs to be clearer? (maybe `getBountiesClaimableByPlayer`?)

Page-: No, it makes sense, I just didn't click the difference

PR #219, opened 2017-12-12
Dumping cargo at 0 exp damages armour

hemberger:
* When at 0 experience, dumping goods does armour damage, but
  will not let you dump if it would take you below 1 armour
  (i.e. you can't destroy your ship dumping cargo).

* Armour damage will be holds/5 (e.g. you can dump 4x 300 holds
  in an IST before you need to UNO). This number may need to be
  adjusted based on feedback.

* There is no rollover. You either lose exp (down to 0) or armour.

* When losing experience, don't show that you lost more experience
  than you have.

The official motivation for this change is to ensure that there
is a tradeoff when slaving. When the slave has exp, it can be lost;
but when you slave at 0 exp, there is no longer any tradeoff.
You basically get slaving for free, which was not intended.
Incurring a ship damage penalty seems like a fitting solution.

More importantly, this idea (suggested by Page) has too perfect of
a role-playing payoff to pass up.

-------------

Depends on PRs #217 and #198.

### DIFF
--- a/engine/Default/cargo_dump_processing.php
+++ b/engine/Default/cargo_dump_processing.php
@@ -10,7 +10,7 @@ if (!is_numeric($amount)) {
 }
 
 if ($amount <= 0) {
-	create_error('You must actually enter an ammount > 0!');
+	create_error('You must actually enter an amount > 0!');
 }
 
 if ($player->isLandedOnPlanet()) {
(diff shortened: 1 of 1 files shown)


hemberger:
PR dependencies have been merged into master.

engine/Default/cargo_dump_processing.php
```diff
@@ -31,39 +31,53 @@
...
-if(is_object($good_distance)) {
-	$good_distance = $good_distance->getRelativeDistance();
+	$container['msg'] = 'You have jettisoned <span class="yellow">'.$amount.'</span> ' . pluralise('unit', $amount) . ' of '.$good_name.' and have lost <span class="exp">'.$lost_xp.'</span> experience.';
+	// log action
+	$account->log(LOG_TYPE_TRADING, 'Dumps '.$amount.' of '.$good_name.' and loses '.$lost_xp.' experience', $player->getSectorID());
 }
-$good_distance = max(1,$good_distance);
+else {
+	// No experience to lose, so damage the ship
+	$damage = round($amount/5);
```
Page-: This should maybe be `ceiling` or have a special message when you take 0 damage (eg 1-2 goods jettisoned) - I'm not fussed though just noticed it

hemberger: Okay, I changed the `round` to a `ceil`.

hemberger:
For reference, the new message looks like this:
image

PR #226, opened 2017-12-15
Add some convenience methods to SmrPlayer

hemberger:
* `getSharingPlayers`: Factors out the logic that interacts with `account_shares_info`. Needed for chat functions that allow displaying info about multiple players.
* `getTurnsGained`: Computes the turns gained since last player update. Generalizes this formula for any given time with optional cache bypass. This is used by `SmrPlayer::updateTurns`, but also all of the chat functions that display info about turns.

### DIFF
--- a/lib/Default/SmrPlayer.class.inc
+++ b/lib/Default/SmrPlayer.class.inc
@@ -217,8 +217,36 @@ class SmrPlayer extends AbstractSmrPlayer {
 		else {
 			throw new PlayerNotFoundException('Invalid accountID: '.$accountIDOrResultArray . ' OR gameID:'.$gameID);
 		}
+	}
 
+	// Get array of players whose info can be accessed by this player.
+	// Skips players who are not in the same alliance as this player.
+	public function getSharingPlayers($forceUpdate=false) {
+		$results = array($this);
 
+		// Only return this player if not in an alliance
+		if (!$this->hasAlliance()) {
+			return $results;
+		}
+
+		// Get other players who are sharing info for this game.
+		// NOTE: game_id=0 means that player shares info for all games.
+		$this->db->query('SELECT from_account_id FROM account_shares_info WHERE to_account_id=' . $this->db->escapeNumber($this->getAccountID()) . ' AND (game_id=0 OR game_id=' . $this->db->escapeNumber($this->getGameID()) . ')');
+		while ($this->db->nextRecord()) {
+			try {
+				$otherPlayer = SmrPlayer::getPlayer($this->db->getInt('from_account_id'),
+				                                    $this->getGameID(), $forceUpdate);
+			} catch (PlayerNotFoundException $e) {
+				// Skip players that have not joined this game
+				continue;
+			}
+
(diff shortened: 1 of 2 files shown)


lib/Default/SmrPlayer.class.inc
```diff
@@ -217,8 +217,35 @@ class SmrPlayer extends AbstractSmrPlayer {
 		else {
 			throw new PlayerNotFoundException('Invalid accountID: '.$accountIDOrResultArray . ' OR gameID:'.$gameID);
 		}
+	}
+
+	// Get array of players who are sharing info with this player
+	// Only returns players who are in the same alliance
```
Page-: How do you get info about players who are sharing with you from another alliance? In general though I would expect this to be a parameter so you can choose to get just the same alliance or not

tools/discord/commands/turns.php
```diff
@@ -31,23 +30,11 @@ function get_turns($player) {
...
-		} catch (PlayerNotFoundException $e) {
-			// Skip players that have not joined this game
-			continue;
-		}
-
-		// players must be in the same alliance
-		if ($player->hasAlliance() && $player->getAllianceID() == $otherPlayer->getAllianceID()) {
-			$results[] = get_turns($otherPlayer);
-		}
+	foreach ($player->getSharingPlayers(true) as $otherPlayer) {
```
Page-: This could be
```php
$results = array_map('get_turns_message', $player->getSharingPlayers(true));
```
particularly if `getSharingPlayers` also returns the current player - it's not necessary, it's just I've become a real fan of functional programming in the past few years

lib/Default/SmrPlayer.class.inc
```diff
@@ -217,8 +217,35 @@ class SmrPlayer extends AbstractSmrPlayer {
...
+	public function getSharingPlayers($forceUpdate=false) {
+		$results = array();
```
Page-: If you use
```php
$results = array($this);
```
then code that uses this automatically has access to all of the players that they can get info about, rather than having to special case getting info about the current player

lib/Default/SmrPlayer.class.inc
hemberger: In the current context, there is no circumstance under which you will share information between players who are not in the same alliance. Should the need arise in the future, such a parameter could be added, but I don't see any value gained in adding one now.

lib/Default/SmrPlayer.class.inc
hemberger: I'm happy to do it this way, since all current uses of this function do indeed include the current player. Do you think the function name needs to change to reflect this?

tools/discord/commands/turns.php
hemberger: Yep, this looks good to me. I'll make this change.

lib/Default/SmrPlayer.class.inc
Page-: I think the name is ok

hemberger:
The requested changes have been made.

PR #227, opened 2017-12-15
Add Dockerfile for Caretaker and NPC's

hemberger:


hemberger:
Yep, multi-stage builds would be very nice. However, the differences between the layers are sometimes expensive (like installing DiscordPHP and Runkit), so while the source code is still being actively developed, having to rebuild so many things because of a single change to the code is too time consuming.

We could perhaps have separate development and production images, but that might be more trouble than it's worth.

PR #231, opened 2017-12-27
Even more discord commands

hemberger:
Adds 8ball and forces commands.

### DIFF
--- a/tools/chat_helpers/channel_msg_8ball.php
+++ b/tools/chat_helpers/channel_msg_8ball.php
@@ -1,0 +1,30 @@
+<?php
+
+function shared_channel_msg_8ball() {
+	static $answers = array(
+		'Signs point to yes.',
+		'Yes.',
+		'Reply hazy, try again.',
+		'Without a doubt.',
+		'My sources say no.',
+		'As I see it, yes.',
+		'You may rely on it.',
+		'Concentrate and ask again.',
+		'Outlook not so good.',
+		'It is decidedly so.',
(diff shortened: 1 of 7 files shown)


tools/chat_helpers/channel_msg_8ball.php
```diff
@@ -0,0 +1,30 @@
...
+		'It is certain.',
+		'Cannot predict now.',
+		'Most likely.',
+		'Ask again later.',
+		'My reply is no.',
+		'Outlook good.',
+		'Don\'t count on it.'
+	);
+
+	return $answers[rand(0, count($answers) - 1)];
```
Page-: I was going to suggest `array_rand` for this, but apparently it has dodgy randomness - wtf php

tools/chat_helpers/channel_msg_forces.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+require_once(TOOLS . 'chat_helpers/channel_msg_seedlist.php');
+
+function shared_channel_msg_forces($player, $option) {
+	$db = new SmrMySqlDatabase();
+	if (!empty($option) && $option == "seedlist") {
```
Page-: If you did
```
if (empty($option)) {
	...
}
else if ($option == 'seedlist') {
	...
}
else {
	...
}
```
then you could avoid having to check the !empty case twice

hemberger: Thanks for the suggestion. I've made this change.

PR #232, opened 2017-12-28
Rename Computer to Computers

hemberger:
This was the only good named in the singular, and so it was
always grammatically incorrect.

Requires a database migration.


hemberger:
This change does not impact MGU. I haven't tested any other 3rd party utilities (there are none I have access to).

PR #233, opened 2017-12-28
Add Grandmaster user ranking

hemberger:
Many players have reached Master, so a new rank will give them
something to work towards.

### DIFF
--- a/db/patches/V1_6_61_12__add_user_ranking.sql
+++ b/db/patches/V1_6_61_12__add_user_ranking.sql
@@ -1,0 +1,2 @@
+-- Add a new user rank
+INSERT INTO user_rankings (rank_name) VALUE ('Grandmaster');


hemberger:
The "Grandmaster" user rank comes in at around 250k user score, which currently two players have already reached.
image

db/patches/V1_6_61_12__add_user_ranking.sql
```diff
@@ -0,0 +1,2 @@
+-- Add a new user rank
+INSERT INTO user_rankings (rank_name) VALUE ('Grandmaster');
```
Page-: Can you specify the id - I'm not sure if the auto increment will necessarily be at the right value

hemberger: I certainly can if it will cause a problem if I don't, but it is slightly inconvenient for me because it invalidates all my dev databases. Normally it's not a huge deal, but I made a lot of map/account changes after my pre-migration backup that would be tedious to redo.

When I applied the migration locally it worked fine:
image

Page-: I don't know if it will or not, I just don't necessarily trust the state of the live db is all, so it might need some intervention when it gets deployed (or it might not)

hemberger: This is the table from the live server, so we should be good!
image

PR #238, opened 2017-12-31
Get mysqli working... really.

hemberger:
* Refactor the Classic and 1.2 `Smr*MySqlDatabase.class.inc` so that the mysql secrets are specified in `config/SmrMySqlSecrets.inc`.
* Add a port and socket argument to the `MySqlDatabase` interface, due to a change in how these properties are passed to the mysqli constructor (i.e. they are no longer specified as part of the host argument).


hemberger:
By the way, I tested this on the live server (where we use a socket) and in my dev build (where we don't) and it worked as expected in both places.

PR #239, opened 2018-01-02
Use Grunt to perform CSS/JS minification and cache busting

hemberger:
Replace minified JS/CSS with expanded originals, with the minified and cache-busted versions to be created as part of the build system. Use a multi-stage build to avoid contaminating the PHP environment with the Grunt environment.

If the Grunt build is not performed, then the JS/CSS files are still served as expected (albeit un-minified) with their original filenames.

Hard-code the CSS template/color variants as a global array so that `grunt-cache-bust` can correctly perform the cache busting source code modifications (whereas it cannot if we dynamically construct the CSS URLs).

See the commit messages for more details. (Also, the individual commits provide much more tractable diffs.)


Dockerfile
```diff
@@ -1,3 +1,21 @@
+FROM node:alpine as builder
+
+WORKDIR /smr/
+
+# See https://github.com/hollandben/grunt-cache-bust/issues/236
+RUN npm i --save grunt grunt-contrib-uglify grunt-contrib-cssmin grunt-cache-bust@1.4.1
```
Page-: If you change this to use `--global` then it will be installed globally rather than the working dir, allowing you to drop the `RUN rm -rf node_modules` below, and changing `npx grunt` to just `grunt`

hemberger: I tried that originally, and ran into a bunch of problems before I discovered that global installation is not supported for grunt: https://github.com/gruntjs/grunt-cli/issues/10

Page-: Oh ok, I swear it used to be standard (even recommended?) to install it globally

hemberger: Here's the error I get if I install globally:
```
Step 5/16 : RUN grunt
 ---> Running in 9b5e88c82eb4
grunt-cli: The grunt command line interface (v1.2.0)

Fatal error: Unable to find local grunt.

If you're seeing this message, grunt hasn't been installed locally to
your project. For more information about installing and configuring grunt,
please see the Getting Started guide:

http://gruntjs.com/getting-started
```
Like they said in the Issue thread, we could probably symlink `node_modules` into the local directory, but at that point what are we really gaining?

hemberger: This is my first time using grunt, so I have no idea how it used to be back in the day :)

Page-: Yeah, it's definitely not worth hacking around, I just thought it was an easy change

PR #242, opened 2018-01-02
Minor cleanup

hemberger:
Fix a couple PHP warnings and remove an unused file.

### DIFF
--- a/admin/Default/album_approve.php
+++ b/admin/Default/album_approve.php
@@ -73,7 +73,7 @@ if ($db->nextRecord()) {
 	if (empty($website))
 		$website = 'N/A';
 	$PHP_OUTPUT.=('<tr>');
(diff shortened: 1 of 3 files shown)


admin/Default/album_approve.php
```diff
@@ -73,7 +73,7 @@ function get_album_nick($album_id) {
 	if (empty($website))
 		$website = 'N/A';
 	$PHP_OUTPUT.=('<tr>');
-	$PHP_OUTPUT.=('<td align="right" width="10%" class="bold">Website :</td><td><a href="'.$webpage.'">'.$webpage.'</a></td>');
+	$PHP_OUTPUT.=('<td align="right" width="10%" class="bold">Website :</td><td><a href="'.$website.'">'.$website.'</a></td>');
```
Page-: Can you make a note to check (well, get someone to check :P  ) the websites on the photo album after this change goes through, I don't know if there might be some that need removing

hemberger: Went through the album, and all websites look fine.

PR #243, opened 2018-01-02
Send procedural messages from Federal Clerk

hemberger:
This is a rework of #150, with only a subset of the changes, which have minor differences from the original PR. It focuses on making all procedural kill/death messages be sent by a "Federal Clerk" account ID rather than the killer/victim (in the case of player and force deaths) and rather than "Port Defenses" in the case of port deaths.

Some of the messages have been slightly modified for consistency (e.g. using BBLinks for clarifying the nature of the message), and the layout has been adjusted to better account for the "Player Messages" inbox sharing messages that both do and do not have senders.

These messages from the "Federal Clerk" do not expire for 365 days, at the request of players, so they can keep the kill/death messages for an entire game.

Here is a complete list of all affected messages (the top player message is for context with a message that has a sender):
image


hemberger:
I agree it is a bit clinical, but so long as we're going with the "You were destroyed by / You destroyed" messages, it makes the most sense and fixes some weird message-related issues.

That said, I really like your idea about ports sending more _colorful_ messages, perhaps a taunting message as a Global or to `getAttackersToCredit` if the port isn't busted before the Fed reinforcements arrive.

hemberger:
For this port message idea, t would be nice if the message sender could include the port number so that they're not all just "Port Defenses", but I can't think of a solution off the top of my head. I'll give it some more thought.

As for this PR, I'll simply un-delete `sendMessageFromPortDefenses` for now.

PR #246, opened 2018-01-03
Use list.js to sort personal and alliance force lists

hemberger:
This dramatically improves performance of sorting the force lists,
which gets quite slow once the list is large (as it always is).

We also convert forces_list.php and alliance_forces.php to templates
in the process.

NOTE: in doing this, I believe we lose the ability to secondary sort.
While this was useful, the page loading is so slow that it loses its
convenience (you need to load the page 4 times just to sort by
desc mines with asc expire time, which is typically the only case
where secondary sorting is used). However, secondary sorting can
be re-implemented by adjusting the list.js sort function, if it is
requested.

### DIFF
--- a/engine/Default/alliance_forces.php
+++ b/engine/Default/alliance_forces.php
@@ -8,25 +8,6 @@ $template->assign('PageTopic',$alliance->getAllianceName() . ' (' . $alliance->g
 require_once(get_file_loc('menu.inc'));
 create_alliance_menu($alliance->getAllianceID(),$alliance->getLeaderID());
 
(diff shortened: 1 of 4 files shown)


hemberger:
The first time I tested secondary sorting it worked. Then I tested it in another scenario and it didn't work, and so I assumed that my first test only worked coincidentally. But maybe I was tired or something. I'll have to check again at some point.

hemberger:
Unfortunately it looks like secondary sort does not work by default with our current list.js configuration.

PR #257, opened 2018-01-07
Alliance option to add Discord widget

hemberger:
Adds a "Discord Server" option to the Alliance Stats page, which when set will enable the Discord widget to display on the Message of the Day page.

Also improved descriptions and links for the chat-related Alliance options.

On Alliance Stats page:
image

On Message of the Day page:
image

### DIFF
--- a/templates/Default/engine/Default/alliance_stat.php
+++ b/templates/Default/engine/Default/alliance_stat.php
@@ -16,9 +16,18 @@ if ($CanChangeDescription) { ?>
 
 if ($CanChangeChatChannel) { ?>
 	<tr>
+		<td class="top">Discord Server ID:&nbsp;</td>
+		<td>
+			<input type="text" name="discord_server" size="30" value="<?php echo htmlspecialchars($Alliance->getDiscordServer()); ?>" />&nbsp;
+			<a href="<?php echo WIKI_URL; ?>/chat#alliance-discord-widget" target="_blank">
+				<img src="images/silk/help.png" width="16" height="16" alt="" title="Goto SMR Wiki: Discord Widget"/>
+			</a>
+			<br />Adds the Discord join server widget to the Message of the Day page.
+		</td>
+	<tr>
 		<td class="top noWrap">Discord Channel ID:&nbsp;</td>
 		<td>
-			<input type="text" name="discord" size="30" value="<?php echo htmlspecialchars($Alliance->getDiscordChannel()); ?>" />&nbsp;
+			<input type="text" name="discord_channel" size="30" value="<?php echo htmlspecialchars($Alliance->getDiscordChannel()); ?>" />&nbsp;
 			<a href="<?php echo WIKI_URL; ?>/chat#autopilot-for-alliance-channels" target="_blank">
 				<img src="images/silk/help.png" width="16" height="16" alt="" title="Goto SMR Wiki: Autopilot"/>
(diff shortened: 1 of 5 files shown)


hemberger:
The only alternative I could think of was to allow html in the message of the day, but then use something like http://htmlpurifier.org/docs to make sure it doesn't contain malicious code. While this affords the most flexibility, I think the functionality added in this PR provides a more obvious way to achieve the desired benefit (rather than having to know 1. that the motd accepts html and 2. that there is a Discord server widget).

templates/Default/engine/Default/alliance_stat.php
```diff
@@ -16,12 +16,32 @@
 
 if ($CanChangeChatChannel) { ?>
 	<tr>
-		<td class="top">Discord Channel ID:&nbsp;</td>
-		<td><input type="text" name="discord" size="30" value="<?php echo htmlspecialchars($Alliance->getDiscordChannel()); ?>" /></td>
+		<td class="top">Discord Server ID:&nbsp;</td>
+		<td>
+			<input type="text" name="discord_server" size="30" value="<?php echo htmlspecialchars($Alliance->getDiscordServer()); ?>" />
```
Page-: Is a size of 30 enough for discord servers? (I assume it probably is as I think they use short urls, but just in case)

templates/Default/engine/Default/alliance_stat.php
```diff
@@ -16,12 +16,32 @@
...
+			<br />Adds the Discord join server widget to the Message of the Day page.
```
Page-: Could you add a little info on where you get the server url? I'm guessing that it'd be via the invite link or similar but I'm actually not sure, so I expect there'll be a bunch of questions about it

templates/Default/engine/Default/alliance_stat.php
hemberger: Yeah, the ID's are currently 18 characters longs, so 30 should be safe.

templates/Default/engine/Default/alliance_stat.php
hemberger: Sure, do you prefer it to link to an SMR Wiki section about it or to include it in the page?

You go to `Server Settings -> Widget` and then there's a `Server ID` field with a "Copy" button. All you need is the ID, not a URL. Then on the Discord side you have to click "Enable Server Widget" and if you also want it to display a "Connect" button on the widget you need to set an "Instant Invite Channel" from the dropdown menu.

I guess one benefit of a wiki page is that it could provide some image examples.

hemberger:
I actually didn't think about BBCode. Since all you'd need is `[discordserver=12345]` on the user side, perhaps that's the better solution here?

templates/Default/engine/Default/alliance_stat.php
Page-: A wiki page probably makes sense, it'll encourage more wiki use and it allows things like image examples and easy updating

templates/Default/engine/Default/alliance_stat.php
Page-: ok

hemberger:
No, I can't really think of any other place where someone would want to insert a BBCode discord widget. Maybe in a player message to invite someone to your server? But at that point it's probably easier to create a Discord invite link, I think.

PR #258, opened 2018-01-07
planet_ownership.php: add message about losing ownership

hemberger:
When you go to claim a planet, it is often annoying to check to
see if you already own a planet (especially now that we have multis).

With this change, a message will be displayed letting you know that
if you take ownership when you already have a planet, you will lose
ownership of that other planet. This implicitly declares the
"one planet per person" rule, which was otherwise not stated anywhere.

Also minor rewording to make the new text fit better in context.

image

image

image

image

image

### DIFF
--- a/engine/Default/planet_ownership.php
+++ b/engine/Default/planet_ownership.php
@@ -15,4 +15,10 @@ $template->assign('ProcessingHREF', SmrSession::getNewHREF($container));
 $template->assign('Planet', $planet);
 $template->assign('Player', $player);
 
+// Check if this player already owns a planet
+$db->query('SELECT sector_id FROM planet WHERE game_id=' . $db->escapeNumber($player->getGameID()) . ' AND owner_id=' . $db->escapeNumber($player->getAccountID()));
+if ($db->nextRecord()) {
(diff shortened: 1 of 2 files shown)


engine/Default/planet_ownership.php
```diff
@@ -15,4 +15,10 @@
 $template->assign('Planet', $planet);
 $template->assign('Player', $player);
 
+// Check if this player already owns a planet
+$db->query('SELECT sector_id FROM planet WHERE game_id=' . $db->escapeNumber($player->getGameID()) . ' AND owner_id=' . $db->escapeNumber($player->getAccountID()));
```
Page-: It might be worth moving this to `$player->getPlanet()`/`$player->hasPlanet()` as it seems like something which would be used elsewhere? (although maybe it isn't, as I guess most places where ownership comes into play is in the planet lists, or from planet -> owner, not owner -> planet)

hemberger: Good idea. Since I added a `SmrPlayer::getPlanet` method in another PR, I'll make a note to come back to this and use it here once that other PR is merged.

PR #259, opened 2018-01-08
Refactor planet list into defense and financial

hemberger:
All "Planets" links now display the same page `planet_list.php`.
This page has been modified to have only a single table:

* Both tables in the `trader_planet.php` page had the same sort
  ID's, because they were included by the same template, so sorting
  on this page was broken. Having a single table fixes this.

* Conversely, on `alliance_planets.php`, where there was already
  only a single table, there was nothing to help identify what
  planet the current player owned.

  So in lieu of the second table exclusively for the player's
  planet, we add a sentence stating which planet the player owns.
  This sentence provides a link to the player's planet within the
  list using an anchor with the class id (`planet-<sector_id>`).

* To continue to support viewing other alliance's planet lists
  (for when planetary treaties are re-enabled), we will only
  include information about the current player if the planet list
  is for the current player's alliance.

We also add a new page `planet_list_financial.php`:

* This page displays all financial info about the planets,
  including credits, interest rate, bond, and bond maturity time.

* The bonds are no longer displayed on `planet_list.php`, both
  in an effort to decrease the table width and to logically
  organize the displayed information.

* Because we now have two planet list pages, these pages have
  a new menu bar, which allows you to select between the "Defense"
  (`planet_list.php`) and "Financial" (`planet_list_financial.php`)
  pages. This can be naturally expanded if/when planets have
  more features that need to be viewed in list form.


hemberger:
Some of the various screens that players can encounter are shown below.

## When you're in an alliance
image

image

image

## When you're not in an alliance
image

image

## On the Financial page
image

image

engine/Default/planet_list.inc
```diff
@@ -0,0 +1,46 @@
+<?php
+
+// The engine files for planet lists have a lot in common, so do
+// most of the work here.
+function planet_list_common($allianceId, $getPlanets) {
+	global $template, $player;
+
+	$playerOnly = $allianceId == 0;
```
Page-: This should be `!$player->hasAlliance()`, that way the method of checking if a player has an alliance or not is contained in the one place and it also allows removing the check below, because you know they can never both be true.  And in fact the uses of `$playerOnly` could probably be replaced by `hasAlliance` calls to make the meaning clearer in the world of the combined planet lists

hemberger: I would really like to replace `$playerOnly` with `!$player->hasAlliance()`, but that would break existing support for hidden players viewing alliance planet lists (since the hidden player might not be in an alliance). I _believe_ that's the only scenario where you'd have a solo player viewing another alliance's planet list.

https://github.com/smrealms/smr/blob/17d7492baa8f55df33eeeb224117e5c32dab90a7/engine/Default/menu.inc#L49

Page-: How does that current code here handle the hidden player case? As far as I can see it would treat them as a normal player

hemberger: If the hidden player is not in an alliance, but they want to view the planet list for alliance X (because apparently hidden players are gods), then even though they will have `$player->hasAlliance() == false` and `$player->getAllianceID() == 0`, the page will have `$allianceId == X`. This will allow the planet list for alliance X to display. Otherwise, it would assume we want to only display the current player's planet.

Page-: Urg, ok, and yeah they are but I've never made use of it (you can't have a hidden player and also play yourself imo)

hemberger: Well, we could get rid of the notion of hidden players, which would certainly simplify something like this. In what context were they used in the past?

Page-: I don't really know, I think it was mostly Azool and @MichaelKunze that used them.  My guess was always that it was a way to look into problems that only occurred for other alliances or as admin things to check if there was anything suspicious going on.  What I personally did in the rare occasion when someone had a bug unique to them for some reason was to use a "login as" type functionality, where I would change my `account_id` in `active_session` to theirs in order to reproduce - it was a pretty uncommon thing to need though, especially after I added the automatic bug reporting (because the majority of stuff is never reported.. and the automatic report can include a load more info as well)

MichaelKunze: Thats something Azool implemented after me. I never used it. Get rid of it. @Page- 's solution works perfectly and did it that way when really needed.

PR #261, opened 2018-01-08
Add includeAllianceID argument to SmrAlliance::getAllianceName

hemberger:
This is the second argument, after `$linked`.

If true, will append `' (' . $this->allianceID . ')'` to the
alliance name. This happens to fix a missing space problem
between the alliance name and ID in the right panel.

Also add this argument to `SmrPlayer::getAllianceName`, which
forwards to the `SmrAlliance` method.

### DIFF
--- a/engine/Default/alliance_broadcast.php
+++ b/engine/Default/alliance_broadcast.php
@@ -1,6 +1,6 @@
 <?php
 $alliance =& SmrAlliance::getAlliance($var['alliance_id'], $player->getGameID());
-$template->assign('PageTopic',$alliance->getAllianceName() . ' (' . $alliance->getAllianceID() . ')');
+$template->assign('PageTopic', $alliance->getAllianceName(false, true));
 require_once(get_file_loc('menu.inc'));
 create_alliance_menu($alliance->getAllianceID(),$alliance->getLeaderID());
 
--- a/engine/Default/alliance_exempt_authorize.php
+++ b/engine/Default/alliance_exempt_authorize.php
@@ -1,6 +1,6 @@
 <?php
 $alliance =& $player->getAlliance();
-$template->assign('PageTopic',$alliance->getAllianceName() . ' (' . $alliance->getAllianceID() . ')');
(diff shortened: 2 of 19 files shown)


templates/Default/engine/Default/includes/RightPanel.inc
```diff
@@ -48,10 +48,7 @@ if(isset($GameID)) { ?>
 	</a>
 	Alignment : <span id="align"><?php echo get_colored_text($ThisPlayer->getAlignment(),number_format($ThisPlayer->getAlignment())); ?></span><br />
 	Alliance : <span id="alliance"><a href="<?php echo Globals::getAllianceHREF($ThisPlayer->getAllianceID()); ?>"><?php
-		echo $ThisPlayer->getAllianceName();
-		if($ThisPlayer->hasAlliance()) {
-			echo '('.number_format($ThisPlayer->getAllianceID()).')';
-		} ?></a></span><br />
+		echo $ThisPlayer->getAllianceName(false, true); ?></a></span><br />
```
Page-: The code in this PR all looks ok, but it does suffer from the problem of a list of boolean vars, which is to say that every time a function like this appears you end up having to recheck what each position means because there's no real meaning to the values at the call site.  PHP doesn't have any great way of solving this that I'm aware of, but the best I do know is to use a dictionary to mimic named parameters like `$ThisPlayer->getAllianceName([ 'linked' => false, 'includeID' => true ])` - I'm ok with merging the code as-is, but it is easier to read/maintain via named params (and in this case because the default `linked` is false it can be shortened to `$ThisPlayer->getAllianceName([ 'includeID' => true ])`)

hemberger: Yes, I agree 100%. The absence of keyword arguments really hurts PHP. There are a lot of functions with 1 or more boolean vars (and even non-boolean vars can be opaque too), so it would be a lot of work to fix. I will merge this as-is, but completely understand and sympathize with the rationale for the suggested change.

PR #263, opened 2018-01-09
album_moderate.php: fix errors

hemberger:
* A major piece of error checking logic was in the wrong order,
  which resulted in displaying a broken profile for moderation.
  We now check the account_id first before trying to display
  any errors.

* Add a "<<Back" button link instead of making the `<h1>` header
  the back button.

* Prefix player name with "Album Nickname" so it's clear what
  is being displayed.

* Fix a few PHP warnings about undefined variables.


hemberger:
image
image
image

PR #264, opened 2018-01-09
Replace absolute paths in links

hemberger:
References to internal content do not need to specify the
webserver URL. So we remove it in favor of relative paths.
Less text means less data to serve!

Also fixes a broken link in the NHA message board threads.


hemberger:
I believe that I verified all links still work.

hemberger:
Agreed. At the very worst, we could make all paths relative to the root with the `/` syntax. Either way, prepending the domain component of the URL is entirely unnecessary.

Page-:
Ah right, I was thinking of URL as eg `/smr/test/` for when smr is under a path rather than at the root, rather than being the bit that includes the full domain/protocol

PR #265, opened 2018-01-09
SmrSession: add some ALWAYS_AVAILABLE pages

hemberger:
* info.php: Whenever I try to use this page, I enter the number
  of players to check, click "Next Page", then enter the name
  and click "Check" and I get a back button error. When I make
  this page ALWAYS_AVAILABLE, the error disappears.

* combat_log_viewer_verify.php: On the live server, when someone
  clicks on a [combatlog=X] BBLink in a _new_ scout message
  (i.e. if their forces are attacked), they get a back button error.

  I am not sure if this will fix the problem for a few reasons:
  1. I cannot reproduce this on my test server.
  2. I have been told (but not seen myself) that this also
     occurs when clicking on the Player BBLink in the same
     scout message, but that page (`message_view.php`) is
     already set as ALWAYS_AVAILABLE.
  3. I don't fully understand how RemainingPageLoads works for
     processing files (which `combat_log_viewer_verify.php` is).

### DIFF
--- a/lib/Default/SmrSession.class.inc
+++ b/lib/Default/SmrSession.class.inc
@@ -2,9 +2,14 @@
 require_once('SmrSessionMySqlDatabase.class.inc');
 if(!defined('USING_AJAX'))
 	define('USING_AJAX',false);
+
 class SmrSession {
 	const ALWAYS_AVAILABLE = 999999;
 	const TIME_BEFORE_EXPIRY = 3600;
+
+	// Defines the number of pages that can be loaded after
+	// this page before the links on this page become invalid
+	// (i.e. before you get a back button error).
 	private static $URL_DEFAULT_REMAINING_PAGE_LOADS = array(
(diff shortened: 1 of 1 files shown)


hemberger:
@Page- If you have any intuition for either of these errors, I would be very eager to hear it.

hemberger:
I'm going to defer this PR until we see if the combatlog BBLink error goes away once we're using the PHP container on the live server. If it does, then this PR can be closed without merging.

hemberger:
Oddly enough, the back button error for the combat logs still occurs in the PHP container on the live server. I didn't expect this, since it should roughly be the same environment.

hemberger:
Page discovered that the issue with the scout messages was different grouping of messages on ajax updates affecting which links were made available.

This is fixed more correctly in https://github.com/smrealms/smr/pull/463.

PR #269, opened 2018-01-09
Add flyway migration Dockerfile

hemberger:
If we want a completely sourceless live server environment, then
we will need to have an image for the database patches.

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -1,0 +1,3 @@
+FROM boxfuse/flyway:latest-alpine
+
+COPY ./patches/ /flyway/sql/


hemberger:
What name do you want me to change, just the commit message?

hemberger:
Okay, I've changed the commit message to be clearer about what the dockerfile is for. I'll also tag it in Docker Cloud with some name that includes "flyway".

hemberger:
I'm just going to include the Dockerfile for now. The current development `docker-compose.yml` for flyway is much more convenient. We'll use this Dockerfile in the `dockerize` repo.

PR #270, opened 2018-01-10
Add tests/phplint.sh

hemberger:
Script to perform basic linting checks on the entire source code.
These checks are by no means sufficient, but they are a necessary
check with complete coverage.

Fixes the following fatal errors:

* As of PHP 5.4, call-time pass by reference was removed.
  Function calls no longer require specifying the `&`.

* hex2bin was added in PHP 5.4. Since we require PHP 5.4+,
  remove the redeclarations of this function.

* universe_create_end.php: fix missing semi-colon

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: php
+php:
+  - '5.6'
+git:
+  depth: false
+script:
+  - tests/phplint.sh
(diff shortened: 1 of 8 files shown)


hemberger:
I've included a Travis-CI build as well to run the phplint.sh test.

PR #272, opened 2018-01-10
Change port raid properties

hemberger:
This change is calibrated to keep the following the same:
 * Level 9 ports (average damage, total defense, loot)
 * Profit per turn for all port levels (as per Page's diagnostics)

While keeping those things fixed, we want to make low level
ports a bit easier. To achieve this, we make the following
tradeoffs for low level ports:
 * Decreased port defenses (fewer base CD's)
    Level 1: 1000 -> 500
 * Slightly increased port offense (higher avg. turret damage)
    Level 1: 150 -> 450
 * Very slightly decreased loot
    Level 1: 4.8 mil -> 4.0 mil
 * Significantly decreased max turret damage
    Level 1: 6 * 250 damage -> 4 * 125 damage

Decreased max damage for lower level ports is important to make
them much less risky (since, e.g., losing a ship is never worth
the ~$4 mil you loot from the port). To achieve this, we make
the following changes to port turrets:
 * All turrets have the same accuracy (10%*Level -> 90%)
 * Number of turrets is now variable (6 -> 3+Level)
 * Turret damage is halved (250 -> 125)

For full details, see:
https://docs.google.com/spreadsheets/d/1LAt0H8G8IHeGhwIIRtpf8gU1_mo3ardzM19kPZ5-4hQ

Cosmetic change:
 * Rename "Port Turret" to "Anti-Theft Sentry Turret"
 * Rename "Planet Turret" to "Zenith Turret" (while we're there)
 * Add a restriction to these weapons so that they can be
   displayed in the weapon list as being restricted to ports/planets.


hemberger:
Though somewhat distinct, the port defense change should be made in conjunction with an alignment change. (In the same PR here primarily for convenience, since they have merge conflicts otherwise.)

Players wanted a more expeditious way to change alignment,
since player kills are unreliable and port raids are
excruciatingly slow (`2*Port Level` for only one raider
for each successful bust).

This patch implements the following 3 changes:

* We need the alignment change to occur for all participating
  players, otherwise there is too much randomness.

* We need the change to occur on every shot so that there is
  some relationship between the number of turns used and
  the size of the alignment change (e.g. we don't want people
  to game the system by coming in for the kill shot).

* We implement a damage threshold for the alignment change
  for two reasons.
    1) With a +/- 1 change on every shot for every player,
       alignment would change too rapidly. So without converting
       alignment to a float and using fractional increments,
       we need some other way to scale everything down.
    2) Without a damage threshold, players could spam a Level 1
       port with a single small weapon to artificially increase
       the number of shots per port to an absurd number. We
       want to ensure the PR's are at least somewhat legitimate.

The threshold is currently set at 300 damage, so that if your
Nuke/HHG hits, you automatically get the alignment change.
This is higher than the typical average damage per shot, but
that is needed to avoid turbo gains. This number should be
tweaked according to player feedback.

htdocs/weapon_list.php
```diff
@@ -174,6 +174,12 @@ function applyFilter() {
 			case 3:
 				echo ('<td align="center" style="color: #06F;">Newbie</td>');
 			break;
+			case 4:
+				echo ('<td align="center" style="color: yellow;">Port</td>');
+			break;
+			case 5:
```
Page-: Could you change these to using named constants please, it would make the code far more readable

lib/Default/AbstractSmrShip.class.inc
```diff
@@ -983,6 +983,19 @@ abstract class AbstractSmrShip {
 		$thisPlayer->increaseExperience(round($results['TotalDamage']/20)); // 1/20 weapon damage to exp.
 		$thisPlayer->increaseHOF($results['TotalDamage'],array('Combat','Port','Damage Done'), HOF_PUBLIC);
 //		$thisPlayer->increaseHOF(1,array('Combat','Port','Shots')); //in SmrPortt::attackedBy()
+
+		// Change alignment if we reach a damage threshold.
+		// Decrease if player and port races are at war; increase otherwise.
```
Page-: It should increase if they're at war, not decrease (the code looks to be doing the right thing but the comment should be fixed to avoid confusion)

htdocs/weapon_list.php
hemberger: Done!

lib/Default/AbstractSmrShip.class.inc
hemberger: Yeah, that was a silly typo. Fixed. :)

PR #273, opened 2018-01-10
SmrPort: improve the port raid news message

hemberger:
Do a little minor wordsmithing to make it flow better, and be
a bit less redundant (re: "federal government").

* Remove the span around the alliance, because you can't change
  the color of BBLink text.

* Add "creds" span around the bounty.

New message:
image
Old message:
image

### DIFF
--- a/lib/Default/SmrPort.class.inc
+++ b/lib/Default/SmrPort.class.inc
@@ -593,9 +593,9 @@ class SmrPort {
 			$this->setReinforceTime($nextReinforce);
 			$this->updateAttackStarted();
 			//add news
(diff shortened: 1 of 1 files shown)


hemberger:
I noticed that since the Alliance BBLink doesn't actually create a link outside the game, this takes away the "yellow" class for alliance names in PR news on the login page, and furthermore "creds" is not defined on the login page. However, I see this as more of an issue with the login page than this message. (Side note: when I did some minor cleanup of the login page, I decided not to include the Default.css because it broke the delicate formatting of the custom login graphics.)

PR #278, opened 2018-01-12
Define constants for buyer restrictions and alignment

hemberger:
This address https://github.com/smrealms/smr/pull/272#discussion_r161069366.


hemberger:
Yeah, there's already a `$player->meetsAlignmentRestriction`, which is halfway between your suggestion and what we currently have here, but I figured it would be better to isolate that change into a separate commit later.

PR #279, opened 2018-01-12
Optimize php-apache image

hemberger:
* Use production php.ini by default
* Disable apache .htaccess

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,15 @@ RUN apt-get update \
 	&& rm -rf /var/lib/apt/lists/* \
 	&& docker-php-ext-install curl json mysqli
 
+# Use the production php.ini unless PHP_DEBUG=1 (defaults to 0)
+ARG PHP_DEBUG=0
+RUN [ "$PHP_DEBUG" = "1" ] && echo "Using development php.ini" || \
+	{ echo "Using production php.ini" \
+		&& mkdir /usr/src/php \
+		&& tar --file /usr/src/php.tar.xz --extract --strip-components=1 --directory /usr/src/php \
+		&& cp /usr/src/php/php.ini-production /usr/local/etc/php/php.ini; \
+	}
+
 # We need to set 'sendmail_path' since php doesn't know about sendmail when it's built
 RUN echo 'sendmail_path = "/usr/sbin/sendmail -t -i"' > /usr/local/etc/php/conf.d/mail.ini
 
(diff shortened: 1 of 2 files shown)


hemberger:
I'm sure there's a lot more fine tuning we can do to both PHP and apache, but it is a veritable jungle of advice out there. I'd be grateful for any bits of wisdom that anyone can impart.

hemberger:
Somewhat related, regarding the `PHP_DEBUG` build arg...

I think a reasonable alternative to this method of splitting debug/production would be to have an `smr-production:latest` build using `smr:latest` as a base image, but I can't see any simple way to tell Docker Cloud to only build `smr-production` _after_ `smr`, so I abandoned that idea. In fact, this is a common problem I face in trying to get various builds to share layers.

This comment is mostly for my records, but always open to feedback.

Dockerfile
```diff
@@ -22,9 +22,21 @@ RUN apt-get update \
 	&& rm -rf /var/lib/apt/lists/* \
 	&& docker-php-ext-install curl json mysql mysqli
 
+# Use the production php.ini unless PHP_DEBUG=1 (defaults to 0)
+ARG PHP_DEBUG=0
+RUN [ "$PHP_DEBUG" = "1" ] || \
+	{ echo "Using production php.ini" \
+		&& mkdir /usr/src/php \
+		&& tar --file /usr/src/php.tar.xz --extract --strip-components=1 --directory /usr/src/php \
```
Page-: Where does `/usr/src/php.tar.xz` come from? Is it there in the base image as I can't see it being added

Dockerfile
```diff
@@ -22,9 +22,21 @@ RUN apt-get update \
 	&& rm -rf /var/lib/apt/lists/* \
 	&& docker-php-ext-install curl json mysql mysqli
 
+# Use the production php.ini unless PHP_DEBUG=1 (defaults to 0)
+ARG PHP_DEBUG=0
+RUN [ "$PHP_DEBUG" = "1" ] || \
+	{ echo "Using production php.ini" \
```
Page-: Should this be `Using development php.ini`?

hemberger: `PHP_DEBUG != 1` if for the live server, so we'll use `php.ini-production` there. `PHP_DEBUG==1` is for development, but I prefer the default values over the values in `php.ini-development`, because it too suppresses in-browser warnings and errors (they go into the logfile, which is much less convenient when debugging).

Dockerfile
hemberger: Yeah, every php image has a tar'd copy of the source code in that location.

Dockerfile
Page-: Oh, you used `||`, I assumed it would be `&&` which is what confused me, eg `if debug is false then use prod` rather than `if debug is true ... else use prod`

hemberger: I originally had the `&&`, but the problem with that is it causes docker to exit on failure if not using production (since you get an overall false return value).

Page-: Ah, what about `[ "$PHP_DEBUG" = "1" ] && echo 'using dev' || echo 'using prod' ...`

hemberger: Sure, that sounds good.

PR #282, opened 2018-01-15
Restructure port data storage to improve serialization

hemberger:
**Serialize without modifying the port**

Break the `goods` member into three more logically distinct
members: `goodIDs`, `goodAmounts`, and `goodDistances`.
The `goodIDs` are broken up into "Buy", "Sell", and "All Goods"
arrays as before, but the latter two are simple assoc. arrays.

We will omit `good{Amounts,Distances}` in the serialization,
and only keep the `goodIDs` (which is all the cached port
users need).

We can also avoid modifying `cachedVersion` by setting this
to true in `__wakeup` instead of `__sleep`. This is because
the port we are serializing is not cached -- only the one
we get when we unserialize.

Code is included in `__wakeup` to temporarily permit unserializing
old port caches (this can be removed after any running game is
over).

(Note: below, a "good" means the lengthy data array that
`Globals::getGood{,s}` returns.)
Due to the modified internal data structure of `SmrPort`, some
of the methods have been changed as well:
* `restockGood` takes a goodID instead of a "good".
* `getLootGoodHREF` now takes a goodID instead of a "good".
* `getGoods` is now private, and is used only to populate
  the `good{IDs,Amounts}` members in the ctor.
  It also throws an exception if called on a cached port
  instead of returning false.
* `getVisibleGoods` is now private, and is just a common
  function for `getVisibleGoods{Sold,Bought}`.
* `getVisibleGoodsAll` has been removed.
* `getVisibleGoods{Bought,Sold}` now return an array of goodIDs
  instead of an array of "goods".
* `getGoods{All,Sold,Bought}` have been replaced with
  `get{All,Sold,Bought}GoodIDs`, and they return an array of goodIDs
  instead of an array of "goods".
* `getGoodStatus` has been renamed `getGoodTransaction` to more
  clearly describe what it does.
* `addCachePorts` no longer needs to store and then reset the
  `$cachedVersion` and `$goods`.
* `__sleep` no longer modifies the instance members, and replaces
  `goods` and `cachedVersion` with `goodIDs` in the serialized
  members.
* `__wakeup` now sets `cachedVersion=true` instead of `__sleep`.

Some unnecessary mysql queries have also been removed from `SmrPort`:
* updating `port_has_goods` amounts in `restockGood` (this is
  redundant with `update` called by `savePorts`?)
* updating `port_has_goods` amounts in `setGoodAmount`.

---------------

**Cache update consistency**

The cached version of the port was only updated for port
upgrades and in-combat downgrades. We now update the cache
for the following events:

* Race change (i.e. port is claimed after a raid)
* All level changes (i.e. after a port is razed)

This should now encompass all scenarios where the cached
version of the port has changed.

We use a `cacheIsValid` member now, and set that to false
in these cases; this allows us to only update the cache
once (in the `update` method).

### DIFF
--- a/engine/Default/shop_goods.php
+++ b/engine/Default/shop_goods.php
@@ -40,14 +40,13 @@ if (!empty($var['traded_xp']) ||
 // test if we are searched. (but only if we hadn't a previous trade here
 }
 elseif ($player->getLastPort() != $player->getSectorID()) {
-	$allGoods = $port->getGoodsAll();
 
 	$base_chance = 15;
-	if(isset($allGoods[5]))
+	if ($port->hasGood(5))
 		$base_chance -= 4;
-	if(isset($allGoods[9]))
+	if ($port->hasGood(9))
 		$base_chance -= 4;
-	if(isset($allGoods[12]))
+	if ($port->hasGood(12))
 		$base_chance -= 4;
 
 	if ($ship->getShipTypeID() == 23 || $ship->getShipTypeID() == 24 || $ship->getShipTypeID() == 25)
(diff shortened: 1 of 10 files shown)


engine/Default/shop_goods.php
```diff
@@ -40,14 +40,13 @@
 // test if we are searched. (but only if we hadn't a previous trade here
 }
 elseif ($player->getLastPort() != $player->getSectorID()) {
-	$allGoods = $port->getGoodsAll();
 
 	$base_chance = 15;
-	if(isset($allGoods[5]))
+	if ($port->hasGood(5))
```
Page-: Changing these to named constants would make the code more readable

hemberger:
The main point is to break it up into properties that are intrinsic to the trade good itself (name, id, basePrice, imageLink, etc.) and properties that are intrinsic to the port (amount, distance). As it stands, the code was entangling the two, making it much more difficult to fix the major bugs. Even worse, because it was stored in an array, you always had to check if the array was "port-aware" (included amount & distance fields) or "port-agnostic".

I'm with you 100% about using objects, though. I already have a patch to add an `SmrTradeGood` class, but I didn't include it here because it made the diff even more overwhelming. I think any awkwardness you are feeling will be obviated at that point (I assume -- please feel free to elaborate on your point if you want to make sure we're on the same page).

Page-:
Ah, I think the difference is how we view the two types of "object", I see it as a "Good Type" which has all the info about the type of good (eg it's name, image, etc) and an instance of that good which includes it's stock and modifiers and other stuff unique to that instance of that good - that said it just uses the name $good pretty much everywhere which doesn't help..

The `SmrTradeGood` class sounds like it will solve the confusion

PR #284, opened 2018-01-16
Remove createPort methods

hemberger:
`SmrPort::createPort` is equivalent to `getPort`, so just
remove it. We can also remove the wrapper with the same name
from `SmrSector`.


hemberger:
Yeah, I was thinking the alternative would be to have a real `createPort` method that fails if the port already exists. But this solution seemed like less trouble. I don't know what is best.

PR #286, opened 2018-01-16
Moving off course does not delete plotted course

hemberger:
The progress you have made will remain until you cancel or reach
your destination (or trigger some other condition that cancels
the plotted course). Jumping will not cancel your course, either.

* If you move to a sector that is not linked to the next sector in
  the course, you will lose the "Follow/Scan Course" links on the
  Current Sector page. When you move back to a sector that is linked
  to the next sector, you will regain those links.

* When off the course, you will have a "Replot Course" button,
  which will replot to the original destination sector, but
  starting from your current sector.

* If you re-enter the course at some later point in the chain, the
  course will be updated to remove all the sectors you skipped.

This is implemented with a simple modification to Plotter.class.inc.
By keeping track of the warp sectors, we now retain enough data to
be able to skip to any point in the plotted path **without**
calculating another expensive `Plotter::findDistanceToX`. See
`Plotter::skipToSector` for the details.

--------------

Also add `Plotter::findReversiblePathToX` to factor out the work of
calling `Plotter::findDistanceToX` so that the resulting path is
reversible.

### DIFF
--- a/engine/Default/course_plot_nearest_processing.php
+++ b/engine/Default/course_plot_nearest_processing.php
@@ -24,14 +24,7 @@ $sector =& $player->getSector();
 if($sector->hasX($realX,$player))
 	create_error('Current sector has what you\'re looking for!');
 
-$path =& Plotter::findDistanceToX($realX, $sector, true, $player, $player);
-if($path===false)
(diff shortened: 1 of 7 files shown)


hemberger:
image

PR #287, opened 2018-01-17
Globals: restructure isValidGame

hemberger:
The previous call chain for many in-game functions was:

getGameType -> isValidGame -> getGameInfo

However, `getGameInfo` was a fairly expensive function, constructing
a large array for every call, only to have `isValidGame` check if
the result is empty or not. The condition for whether or not it was
empty was if the `SmrGame` constructor threw an exception.

To speed up all these calls, we instead do the `SmrGame` constructor
check in `isValidGame`, so `getGameType` calls no longer need to
do the work from `getGameInfo`.

We also simplify the `SmrGame` ctor check by adding the
`GameNotFoundException` class, and throw that instead of the base
`Exception` if the game is not found.

### DIFF
--- a/admin/Default/1.6/universe_create.php
+++ b/admin/Default/1.6/universe_create.php
@@ -15,15 +15,14 @@ $template->assign('EditGameHREF',SmrSession::getNewHREF($container));
 $canEditStartedGames = $account->hasPermission(PERMISSION_EDIT_STARTED_GAMES);
 $template->assign('CanEditStartedGames', $canEditStartedGames);
 
-if($canEditStartedGames) {
-	$games = Globals::getGameInfo();
-}
-else {
-	$games = array();
+$games = array();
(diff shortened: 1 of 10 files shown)


hemberger:
It does have to call `SmrGame::getGame` in both places now, but since that result is cached, it seems a small price to pay for not having to construct that gigantic array in `getGameInfo` just to check if the game is valid (especially since `isGameValid` is called ~20 times on every page).

In fact, `getGameInfo` appears to be mostly a relic from before you restructured (or created?) `SmrGame`, when the only way to get the game info was in this array format.

Page-:
Yeah, it almost certainly is just a relic, there's quite a lot of them around (I'm not sure OO even existed in php when SMR was first written)

hemberger:
How about this for a compromise... I've removed `getPortInfo` entirely :)

PR #288, opened 2018-01-17
SmrAccount: make is{Newbie,Veteran} less expensive

hemberger:
The previous implementation of these methods used `getRank`,
which has to query the user stats in the database. This gets
especially expensive when you're doing it for a list of players:

* Current Players list
* Rankings
* Alliance rosters
* Current Sector Players list

By switching to `maxRankAchieved` for these methods, we can
use data that we already have for the `SmrAccount` (no extra
database calls), which significantly speeds up these pages.

This does mean, however, that there may be a time when a player
has technically reached Fledgeling (the first veteran rank) but
is still reported in italics (Newbie/Beginner). I do not think we
need to be rigorous about this, because it is never reported
inconsistently and the player's `maxRankAchieved` will be updated
next time they log in (or view Rankings page, or Trader Status page).

### DIFF
--- a/lib/Default/AbstractSmrAccount.class.inc
+++ b/lib/Default/AbstractSmrAccount.class.inc
@@ -358,7 +358,9 @@ abstract class AbstractSmrAccount {
 	}
 
 	public function isVeteran() {
-		return $this->isVeteranBumped() || $this->getRank() >= FLEDGLING;
+		// Use maxRankAchieved to avoid a database call to get user stats.
+		// This saves a lot of time on the CPL, Rankings, Rosters, etc.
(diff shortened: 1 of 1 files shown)


hemberger:
It should be updated at the play game screen, too. Isn't that sufficient for all practical purposes?

Page-:
I can't what would trigger an update of it on the play game page?

hemberger:
I believe https://github.com/smrealms/smr/blob/c247a51536c0774755fc119acf04009c51e0292e/engine/Default/game_play.php#L12 should do it, since `getRankName` internally calls `getRank`.

PR #290, opened 2018-01-17
Remove some instances of get_file_loc

hemberger:
For all the class files, we do not need to look for them,
because they are in the same directory. While `get_file_loc`
accounts for a small percent of the runtime, it is an easy
change to make.

Profiling showed `get_file_loc` using 1-2% in an empty sector.

### DIFF
--- a/lib/Default/AbstractSmrPlayer.class.inc
+++ b/lib/Default/AbstractSmrPlayer.class.inc
@@ -4,7 +4,7 @@ require_once('SmrForce.class.inc');
 require_once('SmrGame.class.inc');
 require_once('SmrAlliance.class.inc');
 require_once('missions.inc');
-require_once(get_file_loc('Plotter.class.inc'));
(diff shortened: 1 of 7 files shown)


hemberger:
From Page in Discord:
> This will break semi wars, as semi wars uses a custom SmrAccount.
>
>I think it'd make much more sense to just speed up get_file_loc itself, which will help every place that calls it and the easiest way to do that is to use memoisation (eg https://packagist.org/packages/dominionenterprises/memoize / https://packagist.org/packages/ihor/nspl#user-content-memoizedfunction )
> That way it will only have to do the fs lookups once per filename, and should eliminate almost all cost for the function at the cost of a tiny amount of memory

hemberger:
I think this is too much of a micro-optimization to pursue right now, but something even easier (and perhaps faster too) would be to wrap the requires for the main libraries inside a `class_exists` check.

PR #294, opened 2018-01-17
Fix warnings identified by xdebug

hemberger:

### DIFF
--- a/htdocs/config.php
+++ b/htdocs/config.php
@@ -1,5 +1,5 @@
 <?php
-@date_default_timezone_set(@date_default_timezone_get());
+ini_set('date.timezone', 'UTC');
 
 // Repository paths
 define('ROOT', __DIR__ . '/../');
--- a/lib/Default/Template.class.inc
+++ b/lib/Default/Template.class.inc
@@ -31,7 +31,9 @@ class Template {
 	}
 	
 	public function display($templateName,$forAjax=false, $xmlResponse=false) {
-		@ob_clean();
+		if (ob_get_length() > 0) {
+			ob_clean();
+		}
 		ob_start();
 		$this->includeTemplate($templateName);
 		$this->captures[$this->currentCaptureID].=ob_get_clean();


htdocs/config.php
```diff
@@ -1,5 +1,5 @@
 <?php
-@date_default_timezone_set(@date_default_timezone_get());
+ini_set('date.timezone', 'UTC');
```
Page-: This will change server time from CET (using daylight saving) to always being UTC (no daylight savings), which will change all the in-game time stuff

lib/Default/Template.class.inc
```diff
@@ -31,7 +31,9 @@ class Template {
 	}
 	
 	public function display($templateName,$forAjax=false, $xmlResponse=false) {
-		@ob_clean();
+		if (ob_get_length() > 0) {
+			ob_clean();
```
Page-: Since this appeared, I added this in order to get rid of any junk generated by files before reaching the templates, which is also why none of the files had any whitespace after the closing `?>` - I'd forgotten the reasoning for removing that (which can be tough as linux programs like to enforce an empty line, because a lot of linux programs used to break if it was omitted I believe) - but it would actually make sense to remove the closing tags altogether to avoid that problem and hopefully eventually make it possible to get rid of this call, which is really just a hack to deal with the older code in SMR

htdocs/config.php
hemberger: Indeed. For some reason I checked last night and thought they were on the same timezone. Do you have an objection to using UTC in principle, or just in the middle of a game? (Also, won't this only affect the time display, since `TIME` should be independent of timezone.)

lib/Default/Template.class.inc
hemberger: I'm not sure I understand your recommendation here. Are you suggesting we remove the trailing `?>` from every file and then remove `ob_clean();` here?

htdocs/config.php
Page-: I've got no problem with UTC in principle, it's just that after all these years a lot of players will be used to server time as it currently stands so the change will be a little confusing - and having it in the middle of the game would be even more confusing, so keeping it between games and with an announcement would help a lot.  Also whilst it won't affect game mechanics it will affect the time display in the top left as that's based off of the php timezone right? And what about the time offset in prefs, I don't remember if that is offset against UTC or against server time (I guess realistically it should be timezones rather than a flat offset in order to account for DST, but that might be quite a headache to handle)

lib/Default/Template.class.inc
Page-: Yeah, the trailing `?>` should be removed - check out https://secure.php.net/manual/en/language.basic-syntax.phptags.php
> If a file is pure PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.

As for removing the `ob_clean` the reason we need it at all is because we use output buffering in order to stop
```
?>

```
from sending the trailing newline immediately which was breaking things like redirects (because the redirect is a header and must be sent before any output).  So for that I'd recommend turning it off in dev mode but leaving it on for prod so that it can be found/fixed without breaking prod (like how you enable warnings in dev but not in prod)

hemberger:
This PR is superseded by #401 and #439.

Issue #53, opened 2013-10-16
Changes to Racial Councils


Red-Cross-Knight:
Ref: http://smrcnn.smrealms.de/viewtopic.php?f=94&t=11675

Following changes to be made:
1. Council size = 5
2. Relation change increased from +-10 to +-15
3. Vote length changed to 48 hours (if possible make vote length based on game speed;which would actually put 1.5x at 54 hours)
4. Veto time should = vote time - 24 hours

Further clarification,  please refer to thread or contact Jouldax.

hemberger:
1. Done
2. Effectively done (PR #295 means that relations can change twice as fast)
3. Not implemented (as per Page's comments)
4. Not implemented (hasn't been any further request for this in the past 5 years)

PR #295, opened 2018-01-18
Bilateral Relations

hemberger:
Remove the "our relation with them / their relation with us"
paradigm in favor of a single global relation between two races.

This feature was requested in 2015.

Effect on game dynamics
=======================

Due to the reduced number of players, sometimes races don't have
any active players. This makes it impossible for other races to
ever interact with that race. For the benefit of those still
playing, this change gives them the ability to make slow progress
towards peace with a defunct race.

Since there are 7 other races, which could all in theory be
choosing to change relations with your race, you are no longer
in direct control over your relations. You can still counter some
of them with your votes, and you still have the trump card: a
war vote, which will allow any active council to prevent peace
with another race if they so choose.

I also consider this to satisfy point 2 in issue #53, since a
bilateral relation means that relations can increase twice as
fast. As such, we will keep the relation change at +/-10.

Effect on code performance
==========================

The relations were actually stored (or at least queried) backwards
from the standpoint of performance. Previously, to get your relations
with a race, you would query that race's relations with _all other_
races (see `Globals::getRaceRelations`). So if you needed to get
your relations with all races, you needed to query 8*8=64 relations!

With bilateral relations, all you will ever need to query are the 8
numbers that represent your race's relations with the other races
(and, equivalently, their relations with your race).

The database structure hasn't changed, so we just need to be aware
that querying `Globals::getRaceRelations` with the current player's
race ID is the preferred usage in all normal scenarios.

### DIFF
--- a/engine/Default/council_vote.php
+++ b/engine/Default/council_vote.php
@@ -23,19 +23,17 @@ if ($db->nextRecord()) {
 }
 
 $voteRelations = array();
-$playerRaceGlobalRelations = Globals::getRaceRelations($player->getGameID(),$player->getRaceID());
+$globalRelations = Globals::getRaceRelations($player->getGameID(), $player->getRaceID());
 $races =& Globals::getRaces();
 foreach($races as $raceID => $raceInfo) {
 	if($raceID == RACE_NEUTRAL || $raceID == $player->getRaceID())
 		continue;
 	$container = create_container('council_vote_processing.php', '', array('race_id' => $raceID));
-	$otherRaceGlobalRelations = Globals::getRaceRelations($player->getGameID(),$raceID);
 	$voteRelations[$raceID] = array(
 		'HREF' => SmrSession::getNewHREF($container),
 		'Increased' => $votedForRace == $raceID && $votedFor == 'INC',
 		'Decreased' => $votedForRace == $raceID && $votedFor == 'DEC',
-		'RelationToThem' => $playerRaceGlobalRelations[$raceID],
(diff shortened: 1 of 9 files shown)


Page-:
So it actually used to be the case that you could unilaterally increase another race's relations with your own, which made it possible to force through peace even if the other race did not want it (and war votes require both for the president to not veto it, and for enough players to actively vote) and I'm pretty sure it was a requested change because there were alliances abusing this fact.

hemberger:
I agree that it puts a little more importance on the role of the President (in the specific scenario where one alliance controls the Presidency and Council of Race A and the Presidency of Race B, but the Race B Council doesn't want peace with Race A), but I don't think it creates an inherently abusable situation: first, because it can be considered a reward of political dominance and second, because it is an arbitrary choice to allow the Council (minus the President) to unilaterally keep relations with a race at -500. In the bilateral design, that Council can still drop relations by -40 every day if it is that important to them.

More importantly, this change addresses a very real issue that the current players are facing. The game feels even more empty than it already is when, due to a lack of players, you can't interact with half of the game (racial weapons and ports) because half the races don't have active councils. When the only thing you can do with Politics is give an empty race peace with you (but not peace with them), what's the point? If the Politics design was changed to accommodate the needs of players in the past, then we should be willing to do the same now for the current players.

I'm happy to wait on merging this change until it is approved by the community and admins, but if you'd like to review the implementation of this from a code standpoint, that will help expedite its implementation once approved.

hemberger:
image

PR #296, opened 2018-01-19
Add "Accepted" category to feature requests

hemberger:
This status will help organize the (dauntingly long) list of open
feature requests. Many are good ideas that I plan to implement,
but they continue to clutter the list of open requests that still
need to be dealt with. It will also help give players some quicker
feedback about what is and isn't planned for implementation.

Also reformatted the category display:

* Include a table of categories with descriptions and counts.
* Include the currently viewed category in the table, and
  highlight it in bold.
* Display the currently viewed category in the page heading.

One of the most useful changes is the description of what a "New"
request is. Make the threshold for this category a variable,
and change it from 14 days to 30.

image

### DIFF
--- a/engine/Default/feature_request.php
+++ b/engine/Default/feature_request.php
@@ -10,6 +10,10 @@ if(!isset($var['Status'])) {
 }
 
 $container = $var;
+$container['Status'] = 'Accepted';
+$template->assign('ViewAcceptedFeaturesHref', SmrSession::getNewHREF($container));
+
+$container = $var;
 $container['Status'] = 'Implemented';
 $template->assign('ViewImplementedFeaturesHref',SmrSession::getNewHREF($container));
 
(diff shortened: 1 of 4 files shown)


engine/Default/feature_request.php
```diff
@@ -3,43 +3,44 @@
...
-$container['Status'] = 'Opened';
-$container['ShowOld'] = true;
-$template->assign('ShowOldFeaturesHref',SmrSession::getNewHREF($container));
+$template->assign('PageTopic', 'Feature Requests - ' . $var['category']);
 
-$container = $var;
-$container['Status'] = 'Rejected';
-$template->assign('ShowRejectedFeaturesHref',SmrSession::getNewHREF($container));
+// Feature Requests show up as new for this many days
+$newRequestDays = 30;
```
Page-: This can be defined as a constant as there's nowhere that changes it

hemberger: Done.

PR #304, opened 2018-01-21
 SmrForce: single db query for grouped calls

hemberger:
In the style of `SmrPlayer`, allow passing a db query result
into the `SmrForce` constructor, which then skips the db query
inside the constructor. This allows us to then perform a single
query in `getSectorForces` to avoid the cost of performing a
query for _every_ stack of forces in a sector.

* Modify `getForce` and `SmrForce` ctor to take a
  `gameIDOrResultArray` as the first argument.

* Fix a bug with `forceUpdate` that would have prevented
  `getSectorForces` from updating the cache correctly (but this
  option was not in use, so the bug was only theoretical).

-------------------------
Additionally, remove duplicate db query in `SmrForce` ctor.

Deletion of expired forces is also performed in `tidyUpForces`,
which is called by every public method that gets an `SmrForce`.
But in these methods (unlike in the ctor), the deletion is part
of the cache, which avoids making this call for _every_ stack
of forces!

### DIFF
--- a/lib/Default/SmrForce.class.inc
+++ b/lib/Default/SmrForce.class.inc
@@ -91,7 +91,6 @@ class SmrForce {
 
 	protected function __construct($gameID,$sectorID, $ownerID) {
 		$this->db = new SmrMySqlDatabase();
-		$this->db->query('DELETE FROM sector_has_forces WHERE expire_time < ' . $this->db->escapeNumber(TIME));
 		$this->db->query('SELECT * FROM sector_has_forces
 							WHERE game_id = '.$this->db->escapeNumber($gameID).'
 								AND sector_id = '.$this->db->escapeNumber($sectorID).'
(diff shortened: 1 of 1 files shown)


lib/Default/SmrForce.class.inc
```diff
@@ -89,27 +93,35 @@ class SmrForce {
 		}
 	}
 
-	protected function __construct($gameID,$sectorID, $ownerID) {
+	protected function __construct(&$gameIDOrResultArray, $sectorID, $ownerID) {
 		$this->db = new SmrMySqlDatabase();
-		$this->db->query('DELETE FROM sector_has_forces WHERE expire_time < ' . $this->db->escapeNumber(TIME));
```
Page-: Where is expiration of forces handled now?

hemberger: `SmrForce::getForce` calls `tidyUpForces`
https://github.com/smrealms/smr/blob/9ec5382c97d4a3aa5e8667f297cfa402f9167fee/lib/Default/SmrForce.class.inc#L70
and `tidyUpForces` does the same deletion query that was performed in the ctor
https://github.com/smrealms/smr/blob/9ec5382c97d4a3aa5e8667f297cfa402f9167fee/lib/Default/SmrForce.class.inc#L88

(Note that all calls to the `SmrForce` ctor go through `getForces`.)

PR #305, opened 2018-01-21
SmrShip: only update hardware that has changed

hemberger:
There are up to 11 pieces of hardware at the moment, but often
only one of them is changing at a time (e.g. shields/forces).
However, we previously would update every value if any single
hardware amount changed, resulting in up to 10 unnecessary db
queries.

Because of `set{Old}Hardware`, it is easy to keep track of
which hardware has changed, and only perform the db query to
update those that have changed.

### DIFF
--- a/lib/Default/AbstractSmrShip.class.inc
+++ b/lib/Default/AbstractSmrShip.class.inc
@@ -22,7 +22,7 @@ abstract class AbstractSmrShip {
 
 	protected $hasChangedWeapons = false;
 	protected $hasChangedCargo = false;
-	protected $hasChangedHardware = false;
+	protected $hasChangedHardware = array();
 
 	public static function &getBaseShip($gameTypeID,$shipTypeID,$forceUpdate = false) {
 		if($forceUpdate || !isset(self::$CACHE_BASE_SHIPS[$gameTypeID][$shipTypeID])) {
@@ -334,12 +334,11 @@ abstract class AbstractSmrShip {
 	}
 
 	public function removeAllHardware() {
-		foreach($this->hardware as &$hardware) {
-			$hardware = 0;
+		foreach (array_keys($this->hardware) as $hardwareTypeID) {
+			$this->setHardware($hardwareTypeID, 0);
 		}
 		$this->decloak();
 		$this->disableIllusion();
-		$this->hasChangedHardware = true;
 	}
 
 	public function getPod($isNewbie = false) {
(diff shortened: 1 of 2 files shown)


lib/Default/AbstractSmrShip.class.inc
```diff
@@ -507,7 +506,7 @@ abstract class AbstractSmrShip {
 		if($this->getHardware($hardwareTypeID) == $amount)
 			return;
 		$this->hardware[$hardwareTypeID] = $amount;
-		$this->hasChangedHardware = true;
+		$this->hasChangedHardware[] = $hardwareTypeID;
```
Page-: It would be better to do
```php
$this->hasChangedHardware[$hardwareTypeID] = true
```
that way you guarantee the `$hardwareTypeID` only appears once, with the current method doing
```php
$ship->setHardware(HARDWARE_SHIELDS, 15)
$ship->setHardware(HARDWARE_SHIELDS, 10)
$ship->setHardware(HARDWARE_SHIELDS, 5)
```
would result in 3 database calls at the end (and I wouldn't be surprised at all if combat triggers a `setHardware` for every weapon hit)

hemberger: Good point. I actually had that to begin with, but then I changed it because I didn't want to do the extra check that the value was `true` (even though we probably don't need to do it). I'll change it back. Thanks!

hemberger: Okay, this change has been made.

lib/Default/SmrShip.class.inc
```diff
@@ -120,11 +120,15 @@ class SmrShip extends AbstractSmrShip {
 	}
 	
 	function updateHardware() {
-		if($this->hasChangedHardware === true) {
+		if (!empty($this->hasChangedHardware)) {
 			$this->db->lockTable('ship_has_hardware');
```
Page-: This can probably be removed, with the sector lock in place this should be redundant, and beyond that if the value has already been changed in the background it's not like we refetch it here, so the damage would already have been done and this wouldn't stop it but would likely slow down a load of things

hemberger: Ah yeah, I suppose the sector lock supersedes this hardware lock. But in that case, the only cost incurred by this lock is the expense of `lockTable` and `unlock`, right?

Actually, a more worrying thought just occurred to me. `MySqlDatabase::unlock` unlocks _all_ tables. Does that mean that `updateHardware` (and all other places that call `unlock`) unlocks the sector lock as well?

Page-: They don't touch the sector lock (it's a table itself and doesn't use the mysql lock commands), but the cost of the lock/unlock also includes the cost that no other hardware updates can happen at the same time eg someone uno'ing on one side of the universe will be blocked by someone hitting mines on the other side of the universe

PR #312, opened 2018-01-23
 SmrPlayer: change exp loss/gain on death

hemberger:
* Fix the formulas so that they cannot trigger a DB error:

  - Scale the loss/gain by the max level so that the bounds
    are independent of the absolute magnitude of the level.
    (This was still an issue for Defense World planets, which
    had a much higher max level than other planet types.)

  - As an extra precaution, to ensure that the death is fully
    processed, take the `max(0, $expLoss)`.

* Retool the exp loss/gain percentages:

  - Death to forces: 30% -> 25% (this could be much lower still)
  - Death to ports: [22, 30]% -> [20, 28]%
  - Death to planets: unchanged (as calibrated for level 70)
  - Death to players: [-0.5, 24.5]% -> [5, 25]%
  - Gain on kill: [2, 8]% -> [2.5, 12.5]% (half the loss)

The rationale for the changes are:

1) Force deaths were way too punishing and many people have
   complained about them.
2) Ports and planets are roughly the same, with lower max level
   planets dropping a bit to facilitate the formula based on
   max level. This will allow us to change planet buildings
   without worrying about its impact on this code.
3) A slightly higher average player loss (up to 15% from 12.5%)
   to account for the fact that the average PvP loss is roughly
   half of what it is for any other type of death.
4) Gains are increased slightly due to the much rarer occurrence
   of PvP kills.


hemberger:
Yeah, I was looking at some old HoF stats and I saw games where people were earning 500k+ from kills alone (but that was after many hundreds of kills, with thousands of people playing -- not really a relevant issue for this era).

PR #313, opened 2018-01-23
Fix broken HoF links for pre-1.6 games

hemberger:
The `hall_of_fame_new.php` script only works for 1.6 games. For
the old history databases (pre-1.6 games), this script causes a
database error.

We don't do anything special with this new script, just present the
stats in the simplest way possible.

Current account's stats are highlighted.

----------

NOTE: the "Rankings" db query inserts `$var['stat']`, which is a user-specified column name. This is currently not escaped in any way, so presumably is susceptible to injection. I'm not sure if there is a standard way to escape column names in this code.


hemberger:
image

image

PR #318, opened 2018-01-25
ForceTraderTeamCombatResults.inc: fix kamikaze display

hemberger:
The combat vs. forces results display was incorrect when combat
drones would kamikaze, but also destroy additional forces.

We fix and simplify the display logic here.


hemberger:
Some examples from before this fix:
image
image

And after the fix:
image
image
image

PR #321, opened 2018-01-26
bar_lotto_claim.php: fix claim error

hemberger:
People have been getting the "Multiple actions" error when
claiming a lotto prize.

In a debug environment, claiming a lotto prize resulted in
trying to call `fetch_assoc` on a boolean within `$db->nextRecord()`.
I believe this is because it was a `while`-condition, but within
the `while`-loop, `$db` was modified by additional queries.

Since we know there is only one entry we will be getting, we can
use `if` instead of `while`, which fixes the issue in debug.

What is very strange about the live environment is that the ticket
is deleted, but they don't get the money. The only way this makes
sense to me is if `$player->update()` is not called if a
"Multiple actions" error is encountered. I suppose it must be...


hemberger:
Tested on the live server and it appears to fix the bug.

PR #325, opened 2018-01-26
bar.functions.inc: clean up checkForLottoWinner

hemberger:
* Better document what is going on in this function.
* Simplify the sql queries. We know that we're essentially
  deleting every ticket when a winner is determined, so
  just do it.
* Use a `while` loop for unclaimed prizes instead of `if`
  just in case we somehow get a backlog of unclaimed prizes.

### DIFF
--- a/lib/Default/bar.functions.inc
+++ b/lib/Default/bar.functions.inc
@@ -2,10 +2,11 @@
 
 function checkForLottoWinner($gameID) {
 	global $db;
-	//we we check for a lotto winner...
+	// we check for a lotto winner...
 	$db->lockTable('player_has_ticket');
 	$lottoInfo = getLottoInfo($gameID);
 	$lottoWon = false;
+
 	if ($lottoInfo['TimeRemaining'] <= 0) {
 		//we need to pick a winner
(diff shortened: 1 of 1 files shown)


lib/Default/bar.functions.inc
```diff
@@ -2,33 +2,35 @@
...
 		$db->query('SELECT * FROM player_has_ticket WHERE game_id = '.$db->escapeNumber($gameID).' AND time > 0 ORDER BY rand() LIMIT 1');
 		if ($db->nextRecord()) {
 			$winner_id = $db->getField('account_id');
 			$time = $db->getField('time');
 		}
+
+		// Any unclaimed prizes get merged into this prize
 		$db->query('SELECT * FROM player_has_ticket WHERE time = 0 AND game_id = '.$db->escapeNumber($gameID));
-		if ($db->nextRecord()) {
+		while ($db->nextRecord()) {
```
Page-: You can optimise this by doing
```php
$db->query('SELECT SUM(prize) AS prize FROM player_has_ticket WHERE time = 0 AND game_id = '.$db->escapeNumber($gameID));
if ($db->nextRecord()) {
	$lottoInfo['Prize'] += $db->getField('prize');
}
```

hemberger: Thanks for the suggestion. I've made this change.

PR #329, opened 2018-02-04
 trader_relations.php: visual cleanup

hemberger:
* Convert to an actual table.
* Rename "Global Relations" to "Political Relations" to make it
  clearer what it actually is. (I think internally political or
  "race" relations were called "global" because they were accessed
  via the `Globals` class.)
* Add a new column for the total relations.

image

### DIFF
--- a/engine/Default/trader_relations.php
+++ b/engine/Default/trader_relations.php
@@ -4,37 +4,16 @@ $template->assign('PageTopic','Trader Relations');
 require_once(get_file_loc('menu.inc'));
 create_trader_menu();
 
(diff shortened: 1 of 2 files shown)


Page-:
They were known as "global" prior the to the `Globals` class, I'm not actually sure why they were referred to as that but I was always just used to it from playing

As a side note the Globals class is one of the things which I'd probably not create if I were to do it again, but at the time I was still very heavily influenced by Java's strict OOP where you couldn't have standalone functions and conflated them with globally accessible/mutable state (that I would still avoid though) - that said Globals does the job and can cache db lookups fairly nicely so I don't think it's terrible

PR #330, opened 2018-02-04
SmrPort: swap one good when razing Level 1 port

hemberger:
Previously, you were unable to raze a level 1 port.
Now, it will remove one of the existing T1 goods and
replace it with another.

At some point in the distant past, all level 1 goods
were rearranged when a level was busted. This seems like
a nice compromise between the two.

Note that this only affects razing. Levels still cannot
be busted at level 1 by shot damage.

### DIFF
--- a/lib/Default/SmrPort.class.inc
+++ b/lib/Default/SmrPort.class.inc
@@ -523,15 +523,13 @@ class SmrPort {
 			throw new Exception('Cannot downgrade a cached port!');
 
 		switch($this->level) {
-			case 1: // we don't want to downgrade a level 1 port
-//				for($i=0;$i<3;$i++) { // Add 3 goods for upgrading to level 1
-//					do {
-//						$newGood = array_shift($GOODS);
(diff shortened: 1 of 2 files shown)


lib/Default/SmrPort.class.inc
```diff
@@ -523,15 +523,16 @@ class SmrPort {
...
+				// Replace one class 1 good with another
+				$this->selectAndRemoveGood(1);
+				$newGood = $this->selectAndAddGood(1);
+				// Set new good to 0 supply
+				// (since other goods are set to 0 when port is destroyed)
+				$this->setGoodAmount($newGood['ID'], 0);
+				// We don't want to make the port level 0, so stop here
+				$this->setUpgrade(0);
+				return;
 			break;
```
Page-: You could drop this `break;` since the `return;` means you can never reach it

hemberger: Good point. I slightly changed the logic so that this case doesn't have an alternate return, and instead I wrap the de-leveling call in a `$this->level > 1` conditional.

PR #335, opened 2018-02-07
 planet_financial.php: ajax update financial page

hemberger:
The planet balance and bond can be ajax updated by wrapping them
in a span with a unique id.

The balance is always displayed, so we can span just the number.
However, the bond message is omitted if there is no bond, so we
need to the span around the entire sentence (and outside the
conditional).

### DIFF
--- a/engine/Default/planet_financial_processing.php
+++ b/engine/Default/planet_financial_processing.php
@@ -3,10 +3,10 @@ if (!$player->isLandedOnPlanet())
 	create_error('You are not on a planet!');
 $planet =& $player->getSectorPlanet();
 $action = $_REQUEST['action'];
-$amount = $_REQUEST['amount'];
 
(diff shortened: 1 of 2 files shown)


hemberger:
I can do that, but it looks like I would need/want 3 different spans to cover everything (one around the isClaimed text, one around the bond status, and one around the bond button). Do nested spans work with ajax in the way I would hope? That is, if I put a span around the maturity time (which changes frequently) and then another span around the entire block of text beneath the input elements (which includes the maturity time), will the latter exclude the maturity time update? I'm guessing not...

Alternatively, should I just put a span around the entire block of text (for simplicity) and accept that it will serve more text than it needs to?

Page-:
Nested spans would be included in the diff calc so that wouldn't help, that said it should be possible to add that functionality.  I think the best/easiest option though would be to use multiple spans.

PR #336, opened 2018-02-10
Make changelog more readable

hemberger:
* Increase font size
* Make headers more distinctive (yellow green color)
* Increase spacing between bullets

I realize this can (and should) be done with CSS, but since the
live server is still not dockerized, players will not see CSS
changes right now.

image

### DIFF
--- a/engine/Default/changelog_view.php
+++ b/engine/Default/changelog_view.php
@@ -32,7 +32,7 @@ while ($db->nextRecord()) {
 		$went_live = 'never';
 	}
 
-	$PHP_OUTPUT.=('<b><small>'.$version.' ('.$went_live.'):</small></b>');
+	$PHP_OUTPUT.=('<b>'.$version.' ('.$went_live.'):</b>');
 
 	$PHP_OUTPUT.=('<ul>');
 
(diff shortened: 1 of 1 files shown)


engine/Default/changelog_view.php
```diff
@@ -41,7 +41,7 @@
 				WHERE version_id = ' . $db2->escapeNumber($version_id) . '
 				ORDER BY changelog_id');
 	while ($db2->nextRecord()) {
-		$PHP_OUTPUT.=('<li>' . $db2->getField('change_title') . '<br /><small>' . $db2->getField('change_message') . '</small></li>');
+		$PHP_OUTPUT.=('<li><span style="font-size:125%;color:greenyellow;">' . $db2->getField('change_title') . '</span><br />' . $db2->getField('change_message') . '<br /><br /></li>');
```
Page-: If you use the `lgreen` class it will match the in-game page headers and will only rely on existing things from shared.css

hemberger: I feel like the in-game headers are intentionally a subtle color. For the changelog, I wanted something that stood out a little more since we want to draw people's attention to the item headings (otherwise you just get lost in a wall of text).

This is the same style used by the User Rank display.

PR #338, opened 2018-02-12
newsletter_send_processing.php: swap From and Reply-To

hemberger:
When sending newsletters in bulk, we want the From address to be
consistent (newsletter@smrealms.de) and the Reply-To address to
be unique (bounce_<account_id>@smrealms.de) for the purpose of
unsubscribing from e-mails.

This is recommended by Google's bulk sender guidelines.

### DIFF
--- a/admin/Default/newsletter_send_processing.php
+++ b/admin/Default/newsletter_send_processing.php
@@ -11,7 +11,7 @@ $mail->SMTPKeepAlive	= true;
 
 //$mail->ConfirmReadingTo       = 'newsletter-read@smrealms.de';
 
-$mail->AddReplyTo('newsletter@smrealms.de', 'SMR Support');
+$mail->AddReplyTo('support@smrealms.de', 'SMR Support');
 $mail->Encoding = 'base64';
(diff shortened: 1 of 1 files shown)


admin/Default/newsletter_send_processing.php
```diff
@@ -11,7 +11,7 @@
 
 //$mail->ConfirmReadingTo       = 'newsletter-read@smrealms.de';
 
-$mail->AddReplyTo('newsletter@smrealms.de', 'SMR Support');
+$mail->AddReplyTo('support@smrealms.de', 'SMR Support');
```
Page-: This line should be unnecessary right? Because it's always overridden below?

hemberger: When you send a newsletter to a single recipient, it does not get overwritten.

Page-: Ah ok, it might be worth moving it into that branch then to make the distinction clear

hemberger: Good idea. I've made the change in PR #340

PR #339, opened 2018-02-12
Remove tools/newsletter_send.php

hemberger:
This command-line tool is replaced by the Admin Tool.


Page-:
I seem to remember this was necessary when the script timed out (because you could restart partway through), but I'm guessing server settings were changed to make sure it wasn't timed out - it might be worth checking if the php.ini settings in docker follow the old live settings to allow the admin tools version to finish

hemberger:
Didn't you reset the `php.ini` to the default `php.ini-production` in the upgrade to Ubuntu 16.04? At least, this diff is empty:
```
diff /etc/php/5.6/fpm/php.ini /usr/lib/php/5.6/php.ini-production
```
I'm worried now that we're going to run into major annoyances when we try to send the newsletter in the next couple days. Do you have any recollection of what needs to change to avoid this issue?

PR #357, opened 2018-02-22
Various NPC fixes

hemberger:
* Fix bug with shutting down cleanly
* Fix NPC display name in CPL
* Fix NPC trading due to SmrPort refactoring
* Moderate refactoring

### DIFF
--- a/htdocs/config.inc
+++ b/htdocs/config.inc
@@ -4,9 +4,7 @@ function logException(Exception $e) {
 	$errorType = 'Error';
 	$message='';
 	$currMySQLError='';
-	if(defined('NPC_SCRIPT')) {
-		$message .= 'Script: '.SCRIPT_ID.EOL.EOL.'-----------'.EOL.EOL;
-	}
+
 	if(is_object($account)) {
 		$message .= 'Login: '.$account->getLogin().EOL.EOL.'-----------'.EOL.EOL.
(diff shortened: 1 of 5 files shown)


htdocs/config.inc
```diff
@@ -37,25 +35,10 @@ function logException(Exception $e) {
 		$message .= 'Trace: ' . $ee->getTraceAsString();
 	}
 
-	if(defined('NPC_SCRIPT')) {
-		global $NPC_LOGIN;
-		try {
-			$db = new SmrMySqlDatabase();
-			$db->query('UPDATE npc_logins SET working='.$db->escapeBoolean(false).' WHERE login='.$db->escapeString($NPC_LOGIN['Login']));
```
Page-: If you don't set the npc as not working here, then it will never move or do anything again until manual intervention

hemberger: The job of updating `npc_logins.working` is now relegated to `tools/npc/npc.php`. All exceptions that are thrown are caught in this script, and then the `exitNPC` function will now do what `config.inc` used to do. The only case that isn't covered is when `logException` itself throws an exception, but I think this is fine for now.

For reference, I originally added this "reset" functionality to `npc.php`, and _then_ I discovered that `config.inc` was supposed to be doing this all along, but wasn't due to a misnamed global variable. I removed the code from `config.inc` since it seems more maintainable to try to keep the npc code as self-contained as possible.

Page-: ok

PR #358, opened 2018-02-23
Minor planet attack changes

hemberger:
Gameplay changes:
- Busting a planet with no defenses now costs the normal planet attack turns and sends a message to the owner's alliance (previously it did not cost turns and sent no message).

Cosmetic changes:
- When a planet is busted, the final message to the owner's alliance indicates that the defenses have been breached.
- Minor spacing adjustments and wording changes to the final port/planet attack pages.


hemberger:
image

image

PR #363, opened 2018-02-28
forces_drop_processing.php: don't remove under attack

hemberger:
I don't see any reason why the "Under Attack" warning should be
removed when you are dropping forces. In fact, the warning can be
missed altogether if you are attacked and then drop a force between
the next AJAX update (or page refresh, if AJAX is disabled).

### DIFF
--- a/engine/Default/forces_drop_processing.php
+++ b/engine/Default/forces_drop_processing.php
@@ -207,7 +207,6 @@ $account->log(LOG_TYPE_FORCES, $change_combat_drones.' combat drones, '.$change_
 
 $forces->updateExpire();
 $forces->update(); // Needs to be in db to show up on CS instantly when querying sector forces
-$ship->removeUnderAttack();
 
 forward(create_container('skeleton.php', 'current_sector.php'));
 


Page-:
@hemberger I'm pretty sure this was specifically for the case of dropping CDs - since you lose CDs you get marked as under attack even if you're not (your new CDs are less than your old CDs)

hemberger:
There's an `updateOldAmount` argument to `SmrShip::decreaseCDs`, which is set to `true` in the `forces_drop_processing.php` call. There is no documentation of this argument (yet -- I should add some), but it seems clear that when `true` it suppresses the "under attack" warning.

Page-:
Ok, then it seems fine, it might be that I added that but forgot to then take out the explicit removeUnderAttack

hemberger:
You added a lot of very useful things in a very short period of time, so there were bound to be some oversights. I'm just glad they exist in the first place, since it makes it a lot easier for me to clean things up. :D

Page-:
Yeah, and I have a strong suspicion a lot of it was done in the early hours of the morning :P

PR #364, opened 2018-02-28
SmrForce: use hasChanged attribute

hemberger:
Avoid unnecessary database calls to update SmrForce instances by
utilizing the hasChanged attribute, which can have a dramatic
impact on sectors that have a large number of force stacks.

This attribute had been implemented correctly for a long time,
but never actually used in the `update` method.


Page-:
Wow, yeah this would have a huge impact in some cases - I can't believe I missed actually adding it to the update method :|

PR #369, opened 2018-02-28
Add a way to close accounts but let users reopen

hemberger:
Sometimes users will request that their account be closed (usually
to not receive newsletters anymore). However, it would be nice to
allow these players to reopen their accounts on their own should
they choose to return (since the energy barrier for getting an admin
to reopen is probably too high).

We add a new option to `account_edit.php` that allows admins to
close accounts with the "User Request" reason. This is a special
reason that will trigger the `reopen_account.php` page upon login,
where they can simply press a button to re-open their account.


hemberger:
In "Edit Account" admin tool:

image

Displayed for a "User Request" banned account upon logging in:

image

hemberger:
Yeah, I noticed those closures. As you say, it's probably not important, but I was planning to modify them anyway. Thanks!

PR #370, opened 2018-03-02
DRAFT: Integrate multis into main accounts

hemberger:
NOTE: this is a draft PR. I want to get feedback on the design before I commit to implementing it in full. However, in conjunction with dependent PR #368 (not included here for the sake of diff cleanliness), it is a fully functional draft -- you can log into the game, switch to/create multis, and move around.

I have considered a variety of ways to implement this idea; many were non-starters, others were way too dramatic and would require significant reconstruction... this was the first one that appeared to be both simple and effective.

Since this is such a major facet of the game infrastructure, I would really like to get some thoughtful feedback before continuing.

Current commit message follows below.

----------------------------------

The purpose of this is twofold:
* Allow players to navigate their main/multi accounts without
  having to log in and out.
* Synchronize things between the two accounts when it doesn't
  make sense to have two values (i.e. e-mail address, user
  preferences, bans, etc.).

Add a new table `account_link_login`, which associates a `login_id`
(the new primary key replacing `account_id` in the `account` table)
with an `account_id` (the primary key of `account_link_login`).
Each "login" (`account`) can be associated with multiple "accounts"
(`account_link_login`).

This is a little confusing, since ideally the `account` table would
be renamed `login` (or similar), but we are probably a bit too heavily
invested in the original infrastructure to completely change it now.

We provide a way on the "Play Game" page to switch to/create multis,
including an option to link an existing account as a multi.

Adds two helper methods to SmrAccount:
* getLinkedAccountList - finds all accounts associated with login
* getLinkedAccount - returns an account if it is associated with login


hemberger:
image

image

hemberger:
@Page- I was wondering if you had any comments about this draft, since it would take a lot of work to get it into a polished/final state.

witold-cyrek:
The issue I have here is that multiple accounts are required.  I'd love for you to be able to make multiple players (ie characters) per game on single account.  Of course that means you cannot play two on one account unless you move to some sort of stateless session model.  Thus we are stuck with this.  

There is no easy way to get around this.  Optimally you wanna log in different sessions in different browser windows so you can alt-tab and control both at the same time.  Having single signon doesn't really benefit this.  Other than you dont have to remember both sets of credentials.

PR #382, opened 2018-03-21
Allow header to overlay Galaxy Map

hemberger:
When we are scrolling around the Galaxy Map for large galaxies,
it is a pain to scroll up to the top left corner to change the
galaxy or center on a sector. By having the header in a fixed
position on the page instead, it is always available for use.

Originally, I wanted the header to have a solid background that
would overlap (and obscure) the galaxy map, and then a little
jQuery button that would minimize the header. I didn't know how
to implement this, so I settled for a transparent background,
but no ability to minimize the header. This means the galaxy
map is always slightly obscured, and the header is a bit harder
to read when it overlaps the galaxy map, but I think it is
better than obscuring a significant portion of the galaxy map.

Since I was using pure html/css, I also was unable to implement
this in such a way that the header and main body were scalable
based on the player's font size preference. I had to hard-code
the Galaxy Map font size to match the main body margin specified
in pixels. If this could be done in a more automatic way, that
would be preferable.

Also:
* Simplify header
* Fix undefined variable PHP warning

### DIFF
--- a/htdocs/css/shared.css
+++ b/htdocs/css/shared.css
@@ -303,8 +303,19 @@ a.buttonA  {
 	white-space: nowrap;
 }
 
+/* Galaxy Map */
+div.gal_map_header {
+	padding-left: 15px;
+	position: fixed;
+	top: 0;
+	z-index: 100; /*Overlay header on galaxy map*/
+	color: white; /*Ensure text can be seen in all templates*/
+}
+div.gal_map_main {
+	margin-top: 8em; /*Space for fixed-position header relative to font size*/
+}
 
-/* LM */
+/* Local Map */
 table.lmt {
 	border-collapse: separate;
 	border-spacing: 6px;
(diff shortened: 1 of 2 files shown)


hemberger:
When the galaxy map is scrolled away from the top left corner:
image
But at the top of the page there is still a margin:
image

Page-:
You should be able to use https://api.jquery.com/slidetoggle/ eg
```javascript
$('.gal_map_header').slideToggle()
```
to slide it into/out of view - switching from a class to an id would probably be good though

Page-:
For automatically adjusting to the font-size then using the `em` unit is probably what you want (instead of `px`), as em is relative to font size

hemberger:
Thanks for the suggestions! I've changed the units to `em`, which helped a lot. I see how to do the jQuery now, but the transparent background is actually growing on me now, and the toggle only seems critical with an opaque background. I guess we'll see if there's any feedback on the current implementation.

PR #385, opened 2018-03-21
Remove duplicated trade messages

hemberger:
Only construct the trade message in `shop_goods_processing.php`
instead of both `shop_goods.php` and `current_sector.php`.
The wording of the two had also become unsynchronized at some
point.

Also correctly pluralize the nouns (points, units, credits).


hemberger:
image

image

image

PR #386, opened 2018-03-22
 Allow Underground ships to steal goods

hemberger:
Underground ships (Thief / Assassin / DC) can steal goods instead of buying them:
* Stolen goods cost 0 credits
* Maximum exp is earned

Add small chance to get caught stealing goods:
* 1-9% chance to get caught (from Port Level 9 to 1 respectively)
* Fine is `(Port Level + 1) * Ideal Price`
* Lose 5 alignment when caught


hemberger:
Showing the "Steal" option and the low probability failure:

image

PR #387, opened 2018-03-22
Install NBBC with Composer

hemberger:
Use Composer so that we do not need to include the source code
in the smr repo. Update from 1.4.5 to 2.x, which is maintained
by a new group -- the reason for the major version change is
primarily for Composer and PHP7+ support, which breaks backwards
compatibility with 1.x.

TODO: this escapes HTML elements, which is something we (possibly?)
want in user-generated content, but not in procedurally generated
content, where we will likely have <span> etc. Can we just globally
allow <span> to not be escaped (if so, how?), or...???


hemberger:
@Page- I was hoping I could get your advice on this one. The newer version of NBBC (needed for PHP7 compatibility going forward) clearly handles existing html in the message differently than before (_all_ instances of `&`, `>`, and `<` are escaped):
image

However, there is code in `bbifyMessage` that indicates that it is not entirely new behavior. For example, after the message is parsed, we have always unescaped line breaks:
```php
$message = $bbParser->parse($message);
$message = str_replace('&lt;br /&gt;','<br />',$message);
```
If the live code is working correctly, why don't we have this display issue now?

Since the SMR-related BBCode rules are quite simple, would it be easier to forego NBBC altogether? I don't think too many people would miss the functionality of being able to use non-SMR BBCode like `[b]...[/b]`. But if this current issue can be untangled, I have no problems with continuing to use NBBC.

(Note: We obviously don't want players to be able to include html in messages, and we don't allow it now. I haven't yet figured out how html elements are escaped in player messages but not in other messages, even though all are run through `bbifyMessage`.)

hemberger:
From Page in Discord:
> for the nbbc stuff you need to use `$bbParser->setEscapeContent(false)`
> it looks like that is new as I can't find any mention of it in the nbbc parser we currently use
> and I'm pretty sure the special handling we have for line breaks is because we handle html escaping in the places we need to but we also escape the line breaks, so that unescapes them
> and that replace is actually from before bbcode was even introduced

```php
function HTMLEncode($string) {
return $string;
//Page
//if (!$this->allow_ampersand)
//return htmlspecialchars($string);
//else return str_replace(Array('<', '>', '"'),
//Array('&lt;', '&gt;', '&quot;'), $string);
}
```
> looks like I hacked that file
> because there was no setting for disabling it before

PR #399, opened 2018-03-28
Add new options to Local and Galaxy maps

hemberger:
* Hide Allied Forces: toggles on and off displaying alliance forces.
  If off, only your own forces will display on the map.

* Show Seedlist Sectors: toggles on and off display of sectors that
  are in the alliance seedlist.

We implement these options as part of a PHP session, for simplicity,
to avoid unnecessarily modifying the `account` or `player` database
tables. This is because these options are typically not needed to
persist indefinitely. We will need to monitor community reaction to
this behavior, as well as possible performance impact of PHP sessions
to see if we need to migrate these options to the database at a
later time.

Add a few class methods:

* SmrSector::hasPlayerForces
  This is needed to replace `hasFriendlyForces` when using the
  "Hide Allied Forces" option. We could have added a bool option
  to the existing method, but I think a separate method is clearer.

* SmrAlliance:isInSeedlist
  Specifies whether a sector is in the alliance seedlist. This is
  needed for the "Show Seedlist Sectors" option.

* SmrAlliance::getSeedlist
  Returns an array of sector IDs from the alliance seedlist.
  Seedlist is cached so that it can be called without extra expense
  from `isInSeedlist`.

Note about CSS:

We add the `div.currentSeclm_seedlist` CSS class in addition to the
`div.lm_seedlist` because there are (unavoidable?) issues with
overlaying two background images when they are part of two separate
classes in the same element. (i.e. you can't do `<div class="a b">`
and display background images in both). However, you can use a child
`div` (i.e. `<div class="a"><div class="b">...</div></div>`) or use
a list of items in the CSS background image attribute. We chose the
latter for simplicity.

### DIFF
--- a/engine/Default/map_local.php
+++ b/engine/Default/map_local.php
@@ -9,6 +9,22 @@
 if($player->isLandedOnPlanet())
 	create_error('You are on a planet!');
 
+// Create a session to store temporary display options
+session_start();
+
+// Set temporary options
+if ($player->hasAlliance()) {
+	if (isset($_POST['change_settings'])) {
+		$_SESSION['show_seedlist_sectors'] = isset($_POST['show_seedlist_sectors']);
+		$_SESSION['hide_allied_forces'] = isset($_POST['hide_allied_forces']);
+	}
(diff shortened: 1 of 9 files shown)


hemberger:
image

image

PR #402, opened 2018-03-28
SocialLogin.class.inc: update to v2.12 Facebook API

hemberger:
There are no API changes between 2.7 and 2.12 that affect our
usage, so we can upgrade now. Version 2.12 will reach end-of-life
some time in 2020.


Page-:
Well I guess that was easy :P

PR #419, opened 2018-04-05
Auto-update page to current sector when kicked from planet

hemberger:
In `create_error`, an exception is thrown if `USING_AJAX=true`.
This is because the ajax page refreshes don't really know how
to handle when a different page needs to be displayed without
the user clicking a button.

The most important place where this occurs is when a player is
kicked from a planet (by the owner or in a bust). In the past,
this would cause ajax to stop updating due to the exception.
We could instead forward to the current sector page, but this
would display a disjointed page because a) the left panel links
don't have span id's for updating and so you'd still see the
"Planet Main" link, and b) the middle panel won't update
correctly because we already have ajax-enabled child elements,
which disables the full middle panel update.

Instead what we will do is use `Template::addJavascriptForAjax`
(which needs to be made public so it can be used in the `engine`
files as well as the `template` files) to force the page to
load the current sector, circumventing all the ajax issues.

However, we still need to display _some_ page, since the javascript
won't trigger until after the page loads. Our options are:
1) Display the planet page for maximum continuity, but then if
   the player isn't using javascript, they will be able to
   effectively stay on the planet even when kicked off.
2) Return immediately after adding the javascript (which would
   need to be done both in the `engine` and `template` pages),
   but then a blank middle panel will display (which also isn't
   a great thing to display for non-javascript users).
3) Forward to `current_sector.php` after adding the javascript.
   This only requires us to modify the `engine` files, and
   displays only a marginally broken ajax-updated page.

We also continue to call `create_error` as before whenever
`USING_AJAX=false`, since that will correctly forward to the
current sector and display the error message.

Since this logic is somewhat fragile and is needed for all planet
display pages (not processing pages, since those require an
explicit click by the user), this is a good opportunity to refactor
all the common code out of these pages into `planet.inc`.

### DIFF
--- a/engine/Default/planet.inc
+++ b/engine/Default/planet.inc
@@ -7,6 +7,7 @@ if (!$player->isLandedOnPlanet()) {
 		// Auto-click current sector with javascript to avoid issues with ajax
 		// updates when the display page changes.
 		$container = create_container('skeleton.php', 'current_sector.php');
+		$container['msg'] = '<span class="yellow">WARNING</span>: You have been ejected from the planet!';
 		$currentSectorHREF = SmrSession::getNewHREF($container);
 		$template->addJavascriptForAjax('EVAL', "location.href = '$currentSectorHREF'");
 		forward($container);
--- a/engine/Default/planet.inc
+++ b/engine/Default/planet.inc
@@ -1,0 +1,25 @@
+<?php
+// Common code for all the planet display pages
+
+// If not on a planet, forward to current_sector.php
+if (!$player->isLandedOnPlanet()) {
+	if (USING_AJAX) {
+		// Auto-click current sector with javascript to avoid issues with ajax
+		// updates when the display page changes.
+		$container = create_container('skeleton.php', 'current_sector.php');
+		$currentSectorHREF = SmrSession::getNewHREF($container);
+		$template->addJavascriptForAjax('EVAL', "location.href = '$currentSectorHREF'");
+		forward($container);
+	} else {
+		create_error('You are not on a planet!');
+	}
+}
+
+$planet = $player->getSectorPlanet();
+$template->assign('ThisPlanet', $planet);
+$template->assign('PageTopic', 'Planet : '.$planet->getName().' [Sector #'.$player->getSectorID().']');
+
+require_once(get_file_loc('menu.inc'));
+create_planet_menu($planet);
+
+?>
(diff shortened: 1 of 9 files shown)


engine/Default/planet.inc
```diff
@@ -0,0 +1,25 @@
...
+// Common code for all the planet display pages
+
+// If not on a planet, forward to current_sector.php
+if (!$player->isLandedOnPlanet()) {
+	if (USING_AJAX) {
+		// Auto-click current sector with javascript to avoid issues with ajax
+		// updates when the display page changes.
+		$container = create_container('skeleton.php', 'current_sector.php');
+		$currentSectorHREF = SmrSession::getNewHREF($container);
+		$template->addJavascriptForAjax('EVAL', "location.href = '$currentSectorHREF'");
```
Page-: You should escape `'` in `$currentSectorHREF` - it shouldn't be a problem ever but it helps to future proof and defend against some side effects a bug could have

engine/Default/planet.inc
```diff
@@ -0,0 +1,25 @@
+<?php
+// Common code for all the planet display pages
+
+// If not on a planet, forward to current_sector.php
+if (!$player->isLandedOnPlanet()) {
+	if (USING_AJAX) {
+		// Auto-click current sector with javascript to avoid issues with ajax
+		// updates when the display page changes.
+		$container = create_container('skeleton.php', 'current_sector.php');
```
Page-: It's probably worth adding a message to the current sector screen as well saying something about not being on a planet anymore (to help avoid player confusion)

engine/Default/planet.inc
hemberger: Would you do this with something like this?
```php
$currentSectorHREF = str_replace("'", "\'", $currentSectorHREF);
```

Page-: The safest would probably be to use `json_encode` in order to make sure it covers all the cases (eg `\'` would become `\\'` in the str_replace case)

hemberger: I see, so omitting the single quotes entirely (not even replacing them with double quotes)? I would have expected:
```php
$template->addJavascriptForAjax('EVAL', 'location.href = "' . json_encode($currentSectorHREF) . '"');
```

Sorry, I don't really know javascript... :/

Page-: You should read up about JSON @hemberger, MDN is probably a good starting point: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON
In short though it's just a human readable data format that is commonly used nowadays and has parsers/generators in basically every language.  Admittedly this is a slightly special usage of it in that it makes use of the fact that JSON is also valid javascript code that you can eval directly, rather than using the safer `JSON.parse`.  The main reason to do it this way is to use the existing mechanisms rather than adding an extra special `REDIRECT` element which would take just the string

PR #421, opened 2018-04-05
Move warp into sector table

hemberger:
Due to the way we construct `SmrSector`, we end up querying the
`warp` table for every single sector. This essentially doubles the
cost of the `SmrSector` ctor.

On the other hand, adding an extra `warp` field to the `sector`
table adds a negligible cost to the `sector` database query, so we
do this instead.

### DIFF
--- a/admin/Default/1.6/universe_create_warps.php
+++ b/admin/Default/1.6/universe_create_warps.php
@@ -28,6 +28,10 @@ while ($db->nextRecord()) {
 	$warp1 = SmrSector::getSector($var['game_id'], $db->getInt('sector_id'));
 	$warp2 = SmrSector::getSector($var['game_id'], $db->getInt('warp'));
 	if ($warp1->getGalaxyID() == $warp2->getGalaxyID()) {
+		// For warps within the same galaxy, even though there will be two
+		// sectors with warps, we still consider this as "one warp" (pair).
+		// Since we're looping over all sectors, we'll hit this twice for each
+		// same-galaxy warp pair, so only add 0.5 to avoid double counting.
 		$warps[$warp1->getGalaxyID()][$warp2->getGalaxyID()] += 0.5;
 	} else {
 		$warps[$warp1->getGalaxyID()][$warp2->getGalaxyID()]++;
--- a/admin/Default/1.6/universe_create_warps.php
+++ b/admin/Default/1.6/universe_create_warps.php
@@ -23,15 +23,14 @@ foreach ($galaxies as $gal1) {
 }
 
 //get totals
-$db->query('SELECT * FROM warp WHERE game_id='.$db->escapeNumber($var['game_id']));
+$db->query('SELECT sector_id, warp FROM sector WHERE warp != 0 AND game_id='.$db->escapeNumber($var['game_id']));
 while ($db->nextRecord()) {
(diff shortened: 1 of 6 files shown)


admin/Default/1.6/universe_create_warps.php
```diff
@@ -23,15 +23,14 @@
...
-$db->query('SELECT * FROM warp WHERE game_id='.$db->escapeNumber($var['game_id']));
+$db->query('SELECT sector_id, warp FROM sector WHERE warp IS NOT NULL AND game_id='.$db->escapeNumber($var['game_id']));
 while ($db->nextRecord()) {
-	$warp1 = SmrSector::getSector($db->getInt('game_id'), $db->getInt('sector_id_1'));
-	$warp2 = SmrSector::getSector($db->getInt('game_id'), $db->getInt('sector_id_2'));
+	$warp1 = SmrSector::getSector($var['game_id'], $db->getInt('sector_id'));
+	$warp2 = SmrSector::getSector($var['game_id'], $db->getInt('warp'));
 	if ($warp1->getGalaxyID() == $warp2->getGalaxyID()) {
-		$warps[$warp1->getGalaxyID()][$warp2->getGalaxyID()]++;
+		$warps[$warp1->getGalaxyID()][$warp2->getGalaxyID()] += 0.5;
```
Page-: I think you should add a comment as to why this is `+= .5` because it's not immediately obvious and having it explained in a comment could save a lot of time when reading the code

hemberger:
You're right that a single query is not expensive, so it's not an issue on pages like `current_sector.php`. However, when you get to the Local and Galaxy maps, and you're constructing many sectors, doubling that cost becomes very significant (tens to hundreds of milliseconds). Every sector we construct calls `hasWarp` eventually, so even if we delayed the query until that call, we wouldn't be saving any time.

An alternative would be to keep the `warp` table, and just query the entire table once and cache it -- this is actually how I was originally intending to implement this optimization. But when I looked at how `SmrSector` was used and how we accessed the `warp` table throughout the code, it seemed much more natural to keep the warp information as part of the `sector` table. Nothing was using global information about all the warps, only sector-local information (even the Uni Gen was looping over every sector and calling `hasWarp`!).

admin/Default/1.6/universe_create_warps.php
hemberger: Excellent point. I'll do this now.

hemberger:
This code would be simplified if I stored a `0` instead of `NULL` for every sector that doesn't have a warp (like we do for the `battles` field). But since we keep the sectors from past games around, this would increase the database size by ~1MB. I can't decide if this is worth a slightly clearer code.

hemberger:
Actually, it looks like allowing the column to be `NULL` makes the table larger, and is slightly less performant (since it has to check for nullness). According to https://dev.mysql.com/doc/refman/5.5/en/data-size.html:
> Declare columns to be NOT NULL if possible. It makes SQL operations faster, by enabling better use of indexes and eliminating overhead for testing whether each value is NULL. You also save some storage space, one bit per column. If you really need NULL values in your tables, use them. Just avoid the default setting that allows NULL values in every column.

Since it fits better into the current `SmrSector` infrastructure to use `0`, I think I'll switch to using that in the table as well, unless there are objections to that.

PR #428, opened 2018-04-07
preferences.php: reorganize CSS preferences

hemberger:
Display changes:

* Merge the "Enable Default CSS", "Template", and "Colour Scheme"
  options into a single "CSS Template" option, which lists all
  the possible Template/Colour Scheme variants as well as "None".

  The "None" option is the equivalent of Enable Default CSS == No.
  It is now listed as a separate "CSS Template" option because,
  when disabled, it doesn't matter to the user what the
  Template/Colour Scheme are.

* Rename "Change CSS Link" to "Add CSS Link", and reword the help
  text to make it clearer what this option does. Also removed the
  link to the non-existent CSS file example.

These changes were needed for the following reasons:

a) It wasn't clear how all these knobs interacted.
b) Fewer knobs is easier for the user in general.
c) It was annoying to switch between templates because you often
   needed to change the template, save, then go back and change
   the colour scheme.

The storage of the CSS preference and the logic for how those
preferences are used to render the page is unchanged. This is
_only_ a simplification for the sake of a clearer user interface.

### DIFF
--- a/engine/Default/preferences_processing.php
+++ b/engine/Default/preferences_processing.php
@@ -217,11 +217,14 @@ else if ($action == 'Change Size' && is_numeric($_REQUEST['fontsize']) && $_REQU
 }
 else if ($action == 'Change CSS Options') {
 	$account->setCssLink($_REQUEST['csslink']);
-	$account->setDefaultCSSEnabled($_REQUEST['defaultcss']!='No');
-	if(isset($_REQUEST['template']))
-		$account->setTemplate($_REQUEST['template']);
-	if(isset($_REQUEST['colour_scheme'])&&in_array($_REQUEST['colour_scheme'],Globals::getAvailableColourSchemes($account->getTemplate())))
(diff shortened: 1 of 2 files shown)


hemberger:
image
image

Page-:
It looks to be missing a `Default - None` option where it uses the `Default` template but no colour scheme.  That said I'm not sure anyone uses custom colour schemes nowadays for it to matter at all but there used to be at least a few that did

hemberger:
There currently is no `None` colour scheme for the `Default` template. Freon22 explicitly added one for his template.

```
$ ls htdocs/css/Default
Default.css
$ ls htdocs/css/Freon22
Blue.css  ClassicGreen.css  Default.css  None.css  SeaBreeze.css
```

Page-:
Oh, I thought `None` was a special one that just didn't try to load any colour scheme rather than loading an empty file.

Issue #297, opened 2018-01-19
Docker PHP is slower than host by roughly 5x


hemberger:
Test configurations and rough average page load times:

live (nginx + php-fpm): 0.01s
live (nginx + php-apache docker): 0.05s
dev (nginx + php-fpm both docker): 0.04s
dev (php-apache docker): 0.04s

For the live configurations, mysql is on the host and PHP connects via a socket. For the dev configurations, mysql is in docker (with `/var/lib/mysql` in a data volume) and the PHP container connects with the usual networking scheme.

For reference, the live configuration uses the "overlay2" storage driver, while dev uses "aufs".

Possible causes of the slownless:
* Network latency communicating with mysql (unlikely since the socket connection is also slow)
* Network latency _and/or_ filesystem I/O with volume mounts (I guess... but why would they have the same performance degredation?)
* Apache vs. Nginx+FPM (unlikely since both are slow)
* Production php.ini vs. development (tested both, and only about a 10% change in speed, not a 500% change)

I did some profiling with Xdebug (in dev only, haven't set it up on live), and noticed that the mysql calls were the greatest expense by far on every page, which I assumed confirmed my suspicion that it was somehow related to the way we were connecting to mysql within docker.
![image](https://user-images.githubusercontent.com/846186/35163614-3382df50-fcfc-11e7-9163-0dada08ff79c.png)

However, when I came across this bug report (after flailing with google searches trying in a last desperate attempt to understand what was causing this slowness), I realized that all the file includes did seem a bit expensive.

https://github.com/docker-library/php/issues/493

![image](https://user-images.githubusercontent.com/846186/35163746-a5a7ead0-fcfc-11e7-8c0a-dc50aa345006.png)
(Perhaps ~25% just for `require_once` on this page.)

I think any satisfying conclusion involves using Xdebug to profile the speedy live configuration. In the meantime, we can monitor the above bug report to see if anyone confirms that this is an issue with Docker itself.

PR #432, opened 2018-04-09
Convert defines to consts

Page-:
This matches how class constants are defined and allows the constants to be resolved at compile time rather than runtime.


htdocs/config.php
```diff
@@ -5,14 +5,14 @@
...
 // Repository paths
-define('ROOT', __DIR__ . '/../');
-define('LIB', ROOT.'lib/');
-define('ENGINE', ROOT.'engine/');
-define('WWW', ROOT.'htdocs/');
-define('UPLOAD', WWW.'upload/');
-define('ADMIN', ROOT.'admin/');
-define('TOOLS', ROOT.'tools/');
-define('CONFIG', ROOT.'config/');
+const ROOT = '/smr/';
```
hemberger: Hard-coding the `ROOT` directory will break the non-docker instance that is running on port 444 of the live server. I was mainly keeping this up to test the performance degradation in the docker instance, but now that it is resolved (see #297), perhaps there is no longer any reason to keep that running.

Is the reason for the change that `__DIR__` isn't known at compile time? If so, I'd be willing to endure the additional cost of a few `define`'s if it means we get to keep the flexibility of `__DIR__`. What do you think?

Page-: I actually thought this was in a `config.specific.sample.php` still (it popped up in a bunch of specific files I had locally, and I didn't spot this wasn't one) and therefore that having the default be `/smr/` made the most sense as it allowed `const` and also to use the defaults as-is when running via docker.  As for `__DIR__` in a const expr I don't know as I didn't check because of those reasons

hemberger: Do you want me to give it a try and push if `__DIR__` works?

Page-: Yeah that'd be great, otherwise it's probably gonna be a couple days until I can check it out

hemberger:
According to http://php.net/manual/en/language.constants.predefined.php:
> All these "magical" constants are resolved at compile time, unlike regular constants, which are resolved at runtime.

(I think the manual is a little outdated w.r.t. regular constants!)

Tested and `__DIR__` works fine, so I made the change. There was a small bug in `Convert CSS_URLS...` where you forgot to change an instance of the `$CSS_URLS` into `CSS_URLS`, but I fixed and squashed that.

I'll merge this in as soon as the tests pass. Thanks for commits!

Page-:
Where did I miss $CSS_URLS, I grepped for it :S

hemberger:
They were in `lib/Default/AbstractSmrAccount.class.inc`. Maybe you did `git grep "\$CSS_URLS"`, which will usually be a false negative? It's a mistake I often make.

PR #442, opened 2018-04-11
SectorNavigation.inc: make scan text consistent

hemberger:
The "SCAN" links were visually distracting because the vertical
text had very different spacing than the horizontal text.

By adjusting some of the CSS properties, we can make them more
regular.

Old formatting:
image

New formatting:
image


hemberger:
@Page- This uses some newer CSS3 features (least supported being `inline-grid`). But this was such an ordeal to get the `a` element centered (horizontally and vertically) in the `div`, that I don't know how to do this with better supported CSS. Do you think this is a problem?

hemberger:
I went with Page's suggestions from 2018 (!) in https://github.com/smrealms/smr/pull/442#issuecomment-380719333. As of Feb 2021,  caniuse.com reports 97% support for `flex`, but still only 93% for `inline-grid` (both certainly better supported than 3 years ago though!).

PR #445, opened 2018-04-12
Add flagship functionality

hemberger:
Give alliance command the ability to set a member as the alliance
flagship. A flag will be displayed on the Local Map in the sector
that this player is in. The purpose is to help coordinate ships
in alliance operations.

By adding a single `flagship_id` field to the `alliance` table,
we incur no extra cost by querying the flagship on the Local Map.

We add the following convenience methods:

* SmrSector::hasAllianceFlagship
* SmrAlliance::hasFlagship
* SmrAlliance::{set,get}FlagshipID

The flagship can be set in the "Schedule Operations" page.

image

image


hemberger:
I've cleanup up the look of the 16x16 flagship image for the Local Map. Icon design is quite a subtle art!

image

PR #447, opened 2018-04-12
SectorNavigation.inc: fix duplicate class attribute

hemberger:
The Current Sector navigation links had the `class` attribute
specified twice, and the second one (`class="ajax"`) was being
ignored by browsers and phpquery. Since I don't see why these
elements were intending to be ajax enabled to begin with, we
simply remove it to fix the error.


hemberger:
Page suggested that (when this worked) it was implemented so that switching your `account_id` allowed you to watch what another player was doing (for debugging purposes, etc.). Perhaps phpquery behaved differently in the past or the other instance of `class` was not there. He recommended that we remove it for now, as we can always add it back if we ever need to support player watching again (especially since there are likely many other issues that would arise when player watching since it is an edge-case functionality).

PR #451, opened 2018-04-14
RightPanelShip.inc: fix nesting rules violation

hemberger:
We used the `<noscript>` tags around opening and closing `<a>`
tags separately. Since this results in open elements when the
`</noscript>` tag is called, it violates nesting rules.

While this still works correctly in most browsers, it breaks
HTML validators in non-recoverable ways.

To fix this, we essentially need to add some content that will
only display when JavaScript is disabled (rather than trying to
share the same content, which may not be possible without
violating nesting rules).

While we don't expect many (if any) people to have JavaScript
disabled, we do need _some_ link to the weapon reordering page,
so we just append a link with the text "Click to" in front of
"Show/Hide" in this case. It is not ideal, but I think it's the
best we can do without a significantly more complicated solution.

image


hemberger:
@Page- I'm not really happy with this solution, but it's the best I could come up with. If you have any better ideas, please let me know. Thanks!

hemberger:
This is fixed in a more satisfying way in #667.

PR #456, opened 2018-04-15
Make exception logs available in-game and in e-mail

hemberger:
config.inc: restructure how exceptions are handled

* Always send the automatic bug report to both the in-game box
  (so that devs who can't access `bugs@smrealms.de` can still see
  every automatic bug report) and to the `bugs@smrealms.de` e-mail
  so that we have a permanent record of the report.

* If there is neither a `SmrPlayer` nor a `SmrAccount`, send the
  error message to the in-game box with both account ID and game ID
  set to 0.

* When handling exceptions, always redirect the user to the next
  appropriate game page even if the error logging fails.

### DIFF
--- a/admin/Default/box_view.php
+++ b/admin/Default/box_view.php
@@ -35,26 +35,35 @@ else {
 			$messages[$messageID] = array(
 				'ID' => $messageID
 			);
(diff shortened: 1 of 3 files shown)


hemberger:
image

PR #460, opened 2018-04-18
Major enhancements to the SmrPlanet class

hemberger:
There are a lot of changes here, but the commits are in logical units that clearly delineate each change. See the commit messages for (very detailed) explanations of each change.

Updating the database is less error-prone, but still has the opportunity for rare race conditions. Since these changes are only internal to SmrPlanet, we do not address any of the fundamental infrastructure problems with "action at a distance" (modifying objects that are not in your sector).

The constructor is changed to prevent querying the database unnecessarily -- data is now (mostly) fetched on demand.

Fix bugs in the following methods:
* getMaxBuildings - didn't work when the optional argument was unspecified
* getCurrentlyBuilding - caused player exp to display inconsistently
* update - split financial updates from the rest to avoid possible race conditions

Cache the results of the following methods to reduce database queries:
* updateTypeInfo
* setMaxBuildings
* setOptions

Removed the following methods:
* getTotalBuildings
* setMaxAttackers
* setTypeDescription
* setTypeImage
* setTypeName

Simply the following methods:
* getMaxLevel
* getLevel
* getOptions
* set{CDs,Armour,Shields}

### DIFF
--- a/lib/Default/SmrPlanet.class.inc
+++ b/lib/Default/SmrPlanet.class.inc
@@ -64,8 +64,7 @@ class SmrPlanet {
 
 	public static function &getPlanet($gameID,$sectorID,$forceUpdate = false) {
 		if($forceUpdate || !isset(self::$CACHE_PLANETS[$gameID][$sectorID])) {
-			$p = new SmrPlanet($gameID,$sectorID);
-			self::$CACHE_PLANETS[$gameID][$sectorID] = $p;
+			self::$CACHE_PLANETS[$gameID][$sectorID] = new SmrPlanet($gameID, $sectorID);
 		}
 		return self::$CACHE_PLANETS[$gameID][$sectorID];
 	}
@@ -83,11 +82,6 @@ class SmrPlanet {
 				VALUES (' . $db->escapeNumber($gameID) . ', ' . $db->escapeNumber($sectorID) . ', ' . $db->escapeNumber($inhabitableTime) . ', ' . $db->escapeNumber($type) . ')');
 		}
 		return self::getPlanet($gameID,$sectorID,true);
-		//		if(!isset(self::$CACHE_PLANETS[$gameID][$sectorID])) {
-		//			$p = new SmrPlanet($gameID,$sectorID,true);
-		//			self::$CACHE_PLANETS[$gameID][$sectorID] =& $p;
-		//		}
-		//		return self::$CACHE_PLANETS[$gameID][$sectorID];
 	}
 
 	public static function removePlanet($gameID,$sectorID) {
@@ -107,7 +101,7 @@ class SmrPlanet {
 		unset(self::$CACHE_PLANETS[$gameID][$sectorID]);
 	}
 
-	protected function __construct($gameID,$sectorID,$create=false) {
+	protected function __construct($gameID, $sectorID) {
 		$this->db = new SmrMySqlDatabase();
 		$this->SQL = 'game_id = ' . $this->db->escapeNumber($gameID) . ' AND sector_id = ' . $this->db->escapeNumber($sectorID);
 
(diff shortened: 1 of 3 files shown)


lib/Default/SmrPlanet.class.inc
```diff
@@ -641,26 +640,31 @@ class SmrPlanet {
...
-	public function setMaxBuildings() {
-		$this->maxBuildings = array();
-
-		$this->db->query('SELECT construction_id, max_amount FROM planet_can_build WHERE planet_type_id='.$this->getTypeID());
-		while ($this->db->nextRecord()) {
-			$this->maxBuildings[$this->db->getInt('construction_id')] = $this->db->getInt('max_amount');
+	/**
+	 * Populates the `maxBuildings` attribute from the `planet_can_build` database table.
+	 */
+	private function setMaxBuildings() {
```
Page-: Could you rename this to `update` or `fetch` or something - `set` is pretty confusing alongside functions like `setShields` (the same goes for the other `set` functions that just fetch data)

lib/Default/SmrPlanet.class.inc
```diff
@@ -680,7 +678,7 @@ class SmrPlanet {
 		
 		//trim buildings first
 		$this->setMaxBuildings();
-		foreach ($this->buildings as $id => $amt) {
+		foreach ($this->getBuildings() as $id => $amt) {
```
Page-: Strictly speaking this change isn't needed as you run the fetch function on the line before

lib/Default/SmrPlanet.class.inc
```diff
@@ -680,7 +678,7 @@ class SmrPlanet {
 		
 		//trim buildings first
 		$this->setMaxBuildings();
```
Page-: This trimming will be broken now because the max buildings won't be updated anymore if they've been previously fetched

lib/Default/SmrPlanet.class.inc
hemberger: You're 100% right. I was trying to fit into the naming scheme of the class, but it was already inconsistent, so I guess I should take a moment to rename a few things.

I'm thinking:
* calculateInterest -> matureBond (sounds a bit weird, maybe payoutBond?)
* setMaxBuildings -> initializeMaxBuildings
* updateTypeInfo -> initializeTypeInfo
* getOptions -> getMenuOptions
* setOptions -> initializeMenuOptions

Look okay to you?

lib/Default/SmrPlanet.class.inc
hemberger: Oh, maybe this comment is referring to the `getMaxBuildings` call on the next line? If so, you're right, we could either use the bare attribute or omit `setMaxBuildings` (since the getter ensures that it is set).

lib/Default/SmrPlanet.class.inc
hemberger: Indeed, good catch. I think the correct fix is to invalidate the attributes that are related to planet type (max buildings, type info, and menu options) in the `setTypeID` function, which is the only place that we can change the typeID.

hemberger: I should add that an even _better_ fix is probably to have a `SmrPlanetType` class that correctly handles all the attributes related to planet type (again: max buildings, type info, and menu options), and hold an instance of that class in `SmrPlanet`. Then when we need to change the typeID, we just replace this instance with a new one of a different type. But I think this is more of a second-pass enhancement, since this first pass is already pretty major.

lib/Default/SmrPlanet.class.inc
Page-: I think shortening it to `initX` or `fetchX` would be good, you could also have it directly inside the `getX` methods as well which is common in a bunch of other places.  For `calculateInterest` I think it's an ok name, but something like `checkBondMaturity` could be better

lib/Default/SmrPlanet.class.inc
Page-: Yeah, I was referring to the `$this->setMaxBuildings();` which does the same work as swapping in `getBuildings` does

lib/Default/SmrPlanet.class.inc
Page-: Yeah, having a separate class for managing the planet type would be better but I agree it's a separate pass

hemberger:
I think I've addressed all your comments. Thanks for the feedback!

PR #463, opened 2018-04-20
message_view.php: revamp scout message grouping

hemberger:
The previous design had two main problems:

1. It was difficult to see specific individual scout messages.

2. Messages are marked as read as soon as you open the message box.
   However, scout messages were previously grouped according to how
   many read/unread messages you had from a player. If that number
   crossed the `MESSAGE_SCOUT_GROUP_LIMIT` threshold as a result of
   marking them read, then links that were displayed when you first
   opened the page become stale by the time you click on them.
   (The page doesn't change, but the auto refreshes invalidate any
   links that aren't either `ALWAYS_AVAILABLE` or generated in
   subsequent calls to the code.)

   While we could have simply made all links `ALWAYS_AVAILABLE`, the
   more correct solution is to ensure that the generation of links
   doesn't depend on whether or not messages are marked as read.

The new design makes two (possibly large) database queries to
fetch the scout message groups, and then to fetch all the
individual messages. The latter allows us to ensure that all
possible links away from the page are always generated.

We hide all the individual messages inside each scout message group
(grouped by player), which now also indicates how many of the
individual messages are unread. We use javascript to "Show/Hide" the
X most recent individual messages in each group (we use
`MESSAGE_SCOUT_GROUP_LIMIT` for now). You can still click
"Show all Messages" to display a chronological list of scout messages
that isn't grouped by player.

NOTE: ideally we would have made the javascript dynamically
generate the list of individual scout messages when we click on
"Show/Hide" (as is done for the weapon list, so that we don't need
to fetch _every_ scout message since most will go unlooked at);
however, we would then be unable to display whether or not the
messages are unread, since that is only known on the very first page
load, at which point all messages are marked as read.

Initial view:
image

Expanded:
image

### DIFF
--- a/engine/Default/message_view.php
+++ b/engine/Default/message_view.php
@@ -252,15 +252,6 @@ function displayMessage(&$messageBox, $message_id, $receiver_id, $sender_id, $me
 	$senderName = getMessagePlayer($sender_id,$player->getGameID(),$type);
 	if ($senderName instanceof SmrPlayer) {
 		$sender = $senderName;
-		unset($senderName);
-		$replace = explode('?', $message_text);
-		foreach ($replace as $key => $timea) {
-			if ($sender_id > 0 && $timea != '' && ($final = strtotime($timea)) !== false) { //WARNING: Expects PHP 5.1.0 or later
-				$send_acc = $sender->getAccount();
-				$final += ($account->getOffset() * 3600 - $send_acc->getOffset() * 3600);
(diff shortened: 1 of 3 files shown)


hemberger:
@Page- I had requested a review since we discussed the problem in Discord a bit, and because the change uses JS, which I have less of a handle on than the PHP. I plan to merge this pretty soon, but feel free to review even after it's merged if you want to.

PR #468, opened 2018-04-24
Add alliance invitations

hemberger:
Allow alliance members with permission to change the alliance password
to invite players into the alliance.

Add a new alliance option "Invite Player", which links to a page that
displays the current alliance quotas (vet/total), allows you to invite
specific players, and lists all pending invitations.

Players who are invited will receive an automatic player message with
a link to join the alliance (achieved through a new `join_alliance`
BBCode tag). You can optionally attach a personal message to the
invitation (this is omitted if you are blacklisted or mail banned,
but the invitation is still sent).

Library changes:

* Add `doAllianceCheck=true` argument to `SmrAlliance::canJoinAlliance`
  so that we can check if a player can join an alliance before making
  him leave his current alliance. If `false`, this doesn't do the
  `hasAlliance` check.

* `SmrPlayer::joinAlliance` now deletes all pending invitations for
  the player, to prevent alliance hopping.

* `SmrAlliance::canJoinAlliance` now does the validation check instead of
  doing the check only in `alliance_join_processing.php`.

* Return a text error message in `SmrAlliance::canJoinAlliance` if the
  player is already in an alliance instead of returning `false`.

Additional alliance option:
image

"Invite Player" alliance tool:
image

Message sent to invited players:
image

### DIFF
--- a/engine/Default/alliance_invite_player.php
+++ b/engine/Default/alliance_invite_player.php
@@ -1,0 +1,53 @@
+<?php
+
+$alliance = $player->getAlliance();
+$game = $player->getGame();
+
+$template->assign('PageTopic', $alliance->getAllianceName(false, true));
+require_once(get_file_loc('menu.inc'));
+create_alliance_menu($alliance->getAllianceID(), $alliance->getLeaderID());
+
+// Remove any expired invitations
+$db->query('DELETE FROM alliance_invites_player WHERE expires < ' . $db->escapeNumber(TIME));
+
+// Get list of pending invitations
+$pendingInvites = array();
+$db->query('SELECT * FROM alliance_invites_player
+            WHERE game_id = '.$db->escapeNumber($player->getGameID()).'
+              AND alliance_id = '.$db->escapeNumber($alliance->getAllianceID()));
+while ($db->nextRecord()) {
+	$invited = SmrPlayer::getPlayer($db->getInt('account_id'), $player->getGameID());
+	$invitedBy = SmrPlayer::getPlayer($db->getInt('invited_by_id'), $player->getGameID());
+	$pendingInvites[$invited->getAccountID()] = array(
+		'invited' => $invited->getDisplayName(true),
+		'invited_by' => $invitedBy->getDisplayName(),
+		'expires' => format_time($db->getInt('expires') - TIME, true),
+	);
+}
+$template->assign('PendingInvites', $pendingInvites);
+
+// Get list of players eligible to join this alliance.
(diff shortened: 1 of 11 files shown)


lib/Default/SmrPlayer.class.inc
```diff
@@ -1725,37 +1728,46 @@ class SmrPlayer extends AbstractSmrPlayer {
...
+				case MSG_ADMIN:
+					$expires = 86400 * 365;
+				break;
+				case MSG_CASINO:
+					$expires = 86400 * 31;
+				break;
+				default:
+					$expires = 86400 * 7;
+			}
+			$expires += TIME;
```
Page-: It might be nice to have the `$expires` that is passed in just be how long before expiration, eg "3 days", rather than a fixed time as it is currently ("2018-04-29 17:40:00")

engine/Default/alliance_invite_player.php
```diff
@@ -0,0 +1,53 @@
...
+$invitePlayers = array();
+if ($alliance->getNumMembers() < $game->getAllianceMaxPlayers()) {
+	$db->query('SELECT account_id FROM player
+	            WHERE game_id = '.$db->escapeNumber($player->getGameID()).'
+	              AND alliance_id != '.$db->escapeNumber($alliance->getAllianceID()).'
+	              AND npc = '.$db->escapeBoolean(false).'
+	            ORDER BY player_id DESC');
+	while ($db->nextRecord()) {
+		$invitePlayer = SmrPlayer::getPlayer($db->getInt('account_id'), $player->getGameID());
+		if (in_array($invitePlayer->getAccountID(), array_keys($pendingInvites))) {
```
Page-: This means you can't re-invite someone who clears their messages right?  That's probably a feature in general as the message is likely to have been deleted on purpose but something to be aware of

lib/Default/SmrPlayer.class.inc
hemberger: I agree that would be nice, but doing it for this function only would be a bit inconsistent since there are a lot of other functions that have `$expires` options that are fixed times. I'd want to change it globally if at all.

engine/Default/alliance_invite_player.php
hemberger: You could theoretically send them a message with `[join_alliance=X]`, but practically no one is going to do that. I think the easiest thing to do is add a "Resend invitation" button in the "Pending invitations" section. I'll probably leave that for a future enhancement if it seems needed.

PR #483, opened 2018-05-04
Improve "Join Game" processing

hemberger:
Gameplay changes:
* Streamline and update admin message sent to members who are placed in the Newbie Help Alliance.
* Place all "italics" players into the NHA, not just those who have never joined a game before. Not all games provide the same amount of learning experiences, and this should make the NHA more active.

Internal changes:
* Simplify and optimize queries, and replace raw queries with object methods where possible.
* Remove `account_has_stats` table (used for HoF before v1.6, now just a redundant way to count games joined).

Old NHA admin message:
image

New message:
image


hemberger:
I'm merging this to keep the database patches in the right order, but any feedback you have will be applied retroactively.

PR #495, opened 2018-05-08
universe_create_locations.php: don't unset $sector

hemberger:
The `$sector` variable is a special global, and should not be
overwritten (i.e. dont' use it as the temporary object of a for-loop).

This also fixes a PHP warning that `$sector` is unset in `smr.inc`.

Also update the README with a more complete list of special global variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,18 +220,33 @@ $link = SmrSession::getNewHREF($container);
 ## Global variables
 All pages are called with the following variables available (there may be more)
 
+### $db
+This is a global instance of the SmrMySqlDatabase class. It can be used for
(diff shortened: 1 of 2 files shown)


Page-:
Strictly speaking I think `$sector` is only special when you're inside a game, but it's still good to fix

hemberger:
Yep, that's 100% true. But since the Uni Gen can be used while in a game, this was unsetting the global `$sector`! Fortunately, the global `$sector` wasn't being used for anything except to assign `ThisSector` to the template and to do the skeleton assigns. Unsetting it here had no practical impact other than to cause a PHP warning that was cluttering up the logs.

PR #504, opened 2018-05-16
Simplify plotted course management

hemberger:
Updating the plotted course is now only handled inside
`SmrPlayer::getPlottedCourse` rather than in the various
movement scripts. This reduces code redundancy.

Unlike the movement scripts, `getPlottedCourse` cannot know if we just
moved, so we separate the database query cache from the path updating:

Any time we call `getPlottedCourse` from a new sector (or script),
it will check to see if the path needs to be updated (using the
`plottedCourseFrom` class member). In principle, this update check
could be run every time `getPlottedCourse` is called, especially
since it is very inexpensive, but since `getPlottedCourse` can be
called many times in a single script, it seemed better to simply
avoid repeating the work at the cost of storing an int.

### DIFF
--- a/engine/Default/course_plot_result.php
+++ b/engine/Default/course_plot_result.php
@@ -12,11 +12,7 @@ $path->removeStart();
 $next_sector = $path->getNextOnPath();
 
 if ($player->getSector()->isLinked($next_sector)) {
-
-	// save this to db (if we still have something)
-	if ($path->getTotalSectors()>0) {
 		$player->setPlottedCourse($path);
-	}
 
 	if (!$player->isLandedOnPlanet()) {
 		// If the course can immediately be followed, display it on the current sector page
--- a/lib/Default/Plotter.class.inc
+++ b/lib/Default/Plotter.class.inc
@@ -293,12 +293,13 @@ class Distance {
 		return $this->path[0];
 	}
 	
-	public function followPath($isWarp=false) {
-		if($isWarp===false)
-			$this->distance--;
-		else
+	public function followPath() {
+		$nextSectorID = array_shift($this->path);
+		if (in_array($nextSectorID, array_values($this->warpMap))) {
 			$this->numWarps--;
-		return array_shift($this->path);
+		} else {
+			$this->distance--;
+		}
 	}
 	
 	public function removeStart() {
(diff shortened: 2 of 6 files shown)


engine/Default/course_plot_result.php
```diff
@@ -12,11 +12,7 @@
 $next_sector = $path->getNextOnPath();
 
 if ($player->getSector()->isLinked($next_sector)) {
-
-	// save this to db (if we still have something)
-	if ($path->getTotalSectors()>0) {
 		$player->setPlottedCourse($path);
```
Page-: You didn't unindent this when you moved it out of the `if` so it's now doubly indented

lib/Default/Plotter.class.inc
```diff
@@ -293,12 +293,13 @@ class Distance {
 		return $this->path[0];
 	}
 	
-	public function followPath($isWarp=false) {
-		if($isWarp===false)
-			$this->distance--;
-		else
+	public function followPath() {
+		$nextSectorID = array_shift($this->path);
+		if (in_array($nextSectorID, array_values($this->warpMap))) {
```
Page-: I think this can be simplified and made more performant with
```php
isset($this->warpMap[$nextSectorID])

engine/Default/course_plot_result.php
hemberger: Thanks -- fixed!

lib/Default/Plotter.class.inc
hemberger: Unfortunately, `$nextSectorID` is an array value, not an array key, so this won't work. However, I think we could just flip the way the `warpMap` is constructed (and swap any references to keys <-> values), and then your suggestion would work.

Page-: Ah, it'd still make sense to use `in_array($nextSectorID, $this->warpMap)` in preference then

hemberger: Good point, I'll make this change now. Thanks!

I see you added the strict comparison since your original comment. Was there a specific reason why you felt this was needed?

Page-: I think in this case it would probably work without the strict check, but the strict check helps avoid a whole lot of potential weirdness in general so I think it'd just be best practice to use it - rather than potentially having to go through a load of debugging for a really weird error in the future

PR #527, opened 2018-06-02
Allow reselling combat drones

hemberger:
A feature often requested by players is a way to sell combat
drones. Here we add the ability to resell combat drones at
any location that sells combat drones.

We only allow selling combat drones because they are the only
hardware that affects your combat rating (i.e. most similar to
weapons).

The only reason one may even want to consider selling any other
type of hardware is during a ship trade. Rather than allowing
any hardware to be sold (which could be dangerous for e.g. new
players who accidentally sell all their armour and shields),
it would make more sense to simply provide a refund for surplus
hardware when buying a new ship. We may consider this in a
future patch.

Issues in shop_hardware_processing.php regarding attack
warnings:

* Removed `isUnderAttack`, which was never needed when buying
  hardware anyway. Removing it should prevent suppressing some
  attack warnings in certain scenarios (i.e. someone fires on
  you right before you buy hardware -- you'd want to know this,
  but the previous code would not warn you).

* When selling CDs, there may be cases where attack warnings are
  suppressed. This is not ideal, but necessary in the current
  code infrastructure. It will only be suppressed if you are
  selling CDs and the attacker only does CD damage to you.

image


hemberger:
Oh, I should explicitly say that the refund value is currently the same as for weapons (50% of purchase cost).

PR #551, opened 2018-06-22
SmrSector: optimize multi-sector queries

hemberger:
Major optimization of multi-sector queries (i.e. `getGalaxySector`
and `getLocationSectors`) to allow us to perform only a single
query to cache all sectors instead of separate queries for each
sector.

We refactor `getSector` and the ctor to optionally take a `$db`,
which if specified is expected to already have a sector query
ready to extract.

This improves the performance of the Galaxy Map and UniGen.

### DIFF
--- a/htdocs/map_galaxy.php
+++ b/htdocs/map_galaxy.php
@@ -97,6 +97,7 @@ try {
 		}
 	}
 
+	$galaxy->getSectors(); //optimized call to cache all sectors first
 	if (isset($sectorID)) {
 		$template->assign('FocusSector', $sectorID);
 		$mapSectors = $galaxy->getMapSectors($sectorID);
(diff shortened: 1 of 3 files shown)


htdocs/map_galaxy.php
```diff
@@ -97,6 +97,7 @@
 		}
 	}
 
+	$galaxy->getSectors(); //optimized call to cache all sectors first
```
Page-: It would make sense imo to put this in `SmrGalaxy->getMapSectors` when `$dist == false || ($dist * 2 + 1 >= $galaxy-> getHeight() && $dist * 2 + 1 >= $galaxy->getWidth())` so that it will automatically be used whenever it makes sense in both local and galaxy map

hemberger: Yep, I completely agree. My plan was to optimize `getMapSectors` itself so that it would benefit from using a single multi-sector query regardless of the function arguments or galaxy size, in which case the calls to `getSectors` aren't needed. I suppose I should just make your suggested change now in case I don't finish my plan as soon as I expect.

Page-: It's probably worth doing it quickly, you've seen all the bits of code from things I planned to do.. :P

PR #566, opened 2018-07-07
{government,underground}.php: convert to template

hemberger:
Additional changes:

* Rename the HQ's based on their location name so they don't
  all say "Federal Headquarters" (instead, they will now say
  "<Race> Headquarters").

* Rename button for joining HQ from "Become a gang member" to
  "Become a smuggler". Gang member doesn't have the right
  connotation.

* Replace the "Become a ..." form with a submit-style link.
  Therefore, we remove the (redundant) check of the form action
  in government_processing.php.


hemberger:
Requesting @smrealms/game-admins approval for the button text change at the Underground.

PR #572, opened 2018-07-14
Re-implement talking to the bartender

hemberger:
We add some introductory text to the main bar page, which leads
naturally into a "Talk to Bartender" option. On this page, you
can hear random gossip that the bartender has heard, or tell him
some gossip yourself.

Add a method to monitor bartender gossip in the Admin Message Boxes.


hemberger:
image

image

image

PR #579, opened 2018-07-31
DRAFT: hunter wars + new planet type

hemberger:
@Page- This PR is just intended to give you a chance to give feedback on some major infrastructural changes that I'm making in order to facilitate the new game type and (especially) the new planet type.

I may have to merge in some or all of this as we prepare to get ready for the next game, possibly before you get a chance to look it over, but don't let that stop you from providing feedback. I will do my best to address any comments you have, retroactively if necessary.

This isn't the final version of what I intend to push (at the very least it will have a cleanup + documentation pass), but barring any substantial criticisms, the final version will probably look very close to this.

Most major changes:

* Moving all the constant planet databases into PHP arrays. I know we had discussed this in #88, and at the time you requested to leave it in the database in case we ever wanted to make an Admin Tool that would allow admins to modify planets (like in Edit Locations). However, there is no accountability in a tool like that, and it is difficult to tell if the database is the same as what you'd expect in a fresh dev environment (we would have to add another table to track local changes). It is much easier for me to develop with the data stored in class constants, changes are more transparent, and it's also more performant (OPcache vs. query overhead).

* For the Hunter Wars game, I need to override the `SmrPort::MAX_LEVEL` constant, so I moved the current `SmrPort` to `AbstractSmrPort`, and then the different game types can implement their own `SmrPort` (in much the same way that `SmrAccount` was abstracted for Semi Wars).


hemberger:
Superseded by #588 and #590.

PR #581, opened 2018-08-05
Refactor shop_ship.php

hemberger:
Split shop_ship.php into a part that queries the ship list and
a part that processes the list (shop_ship.inc).

This is the easiest way to change the ship list for different
game types, because then the query in shop_ship.php can be modified
without duplicating the logic in shop_ship.inc.

### DIFF
--- a/engine/Default/shop_ship.inc
+++ b/engine/Default/shop_ship.inc
@@ -1,0 +1,31 @@
+<?php
+$template->assign('PageTopic','Ship Dealer');
+
+$shipsSold = array();
+if ($db->getNumRows() > 0 ) {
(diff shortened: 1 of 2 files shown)


Page-:
I assume this is for hunter wars?  If so I think it'd make more sense for this to use `$location->getShipsSold()` and then do the abstract location/instantiated split

hemberger:
Yes, it is needed for Hunter Wars. And yes, that would be better! For some reason I assumed that method didn't exist yet, since it wasn't used in `shop_ship.php`. It's a bit more work, but I will revert this change and use your suggestion.

Page-:
Yeah, I probably added it with the others and never switched this db query over, it wouldn't be the first time :P

PR #582, opened 2018-08-05
Use actual shop names

hemberger:
Replace the generic names used for ship and weapon shops with the actual names of the shop.

### DIFF
--- a/engine/Default/shop_ship.inc
+++ b/engine/Default/shop_ship.inc
@@ -1,5 +1,7 @@
 <?php
-$template->assign('PageTopic','Ship Dealer');
+
+$location = SmrLocation::getLocation($var['LocationID']);
+$template->assign('PageTopic', $location->getName());
 
 $shipsSold = array();
 if ($db->getNumRows() > 0 ) {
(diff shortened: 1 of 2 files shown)


hemberger:
image

engine/Default/shop_ship.inc
```diff
@@ -1,5 +1,7 @@
 <?php
-$template->assign('PageTopic','Ship Dealer');
+
+$location = SmrLocation::getLocation($var['LocationID']);
+$template->assign('PageTopic', $location->getName());
 
 $shipsSold = array();
 if ($db->getNumRows() > 0 ) {
```
Page-: Doesn't this get poisoned by the the `SmrLocation::getLocation` above? (at least if it isn't already cached)

hemberger: I believe `$db` would only be poisoned by `SmrLocation::getLocation` if they were using the same `SmrMySqlDatabase` instance, since the `$dbResult` and `$dbRecord` are not static members. But they're using different instances (one global, and the other internal to `SmrLocation`), so it should be good.

PR #589, opened 2018-08-08
PlanetList.inc: none -> None

hemberger:
Capitalize the word "none" when there are no Supplies on a planet,
for consistency with the "Nothing" when there is no Build.


hemberger:
Thanks for the suggestion, Page!

PR #606, opened 2018-09-03
Clean up weapon display tables

hemberger:
* Shink some columns to reduce empty space
* Add "%" symbol to Accuracy consistently
* Add Power Level to mounted weapons on planets

image

image

image


hemberger:
Weapon shops now with colour:
image

PR #609, opened 2018-09-09
Remove account linking to compatibility databases

hemberger:
Account linking is now closed. Any accounts that remain linked to
Classic or 1.2 account ID's will remain linked, but no new linking
will occur.

This allows us to no longer carry around the massive compatibility
databases, and instead only keep the lightweight history databases.

This removes the following classes:
* Smr12MySqlDatabase
* SmrClassicMySqlDatabase

And the following methods:
* SmrAccount::linkAccount
* SmrAccount::upgradeAccount
* SmrAccount::hasAllOldAccountIDs
* Globals::useCompatibilityDatabases
* Globals::getCompatibilityDatabases


hemberger:
Two additional motivations to stop using these compatibility databases:
* The linking/upgrading code was a bit difficult to maintain as account-related procedures changed.
* The old databases contain plain text passwords and e-mail addresses, so they are more of a security hazard than the live database.

PR #617, opened 2018-10-18
Replace Announcements with Twitter Feed on login page

hemberger:
* Replace "Announcements" query from `game_news` table with
  the @SMRealms twitter feed. We'll keep the `game_news` table
  around for now for historical reasons.

* Remove the "Players Online Now" counter. This is usually not
  a good advertisement for the game, since the number of concurrent
  players isn't a good indication of activity. Perhaps better would
  be the number of sessions from the past 24 hours (but we don't
  have a way to track this yet).

* Re-organize the login page elements to fit better with the
  Twitter feed:
    - Center the login form in its <td> element by adding the
      `center` class (this also requires removing the `margin`
      attribute from the `collapsed` class).
    - Split the "Recent News" table into a separate table from
      the login form and announcements table.

NOTE: For the Twitter iframe, we use webkit to restyle the scrollbar
to fit the SMR color theme, but this is not available for Firefox/IE.
There's a lot of special machinery needed to restyle the content
inside an iframe, and we delay the display of the content until it
is completely restyled to avoid the content shifting around as we
update it.

Here's what the new login page looks like:
image

### DIFF
--- a/htdocs/css/login.css
+++ b/htdocs/css/login.css
@@ -78,7 +78,6 @@ table.center {
 	border-spacing: 0px;
 	border-collapse: collapse;
 	padding: 0px;
-	margin: 0px;
 	font-size: 0px;
 }
 input.uncollapse {
@@ -86,4 +85,8 @@ input.uncollapse {
 }
 table.nobord {
 	border: 0px;
+}
+.hide {
+	/* hide element but preserve page layout (note: "visibility" didn't work) */
+	opacity: 0;
 }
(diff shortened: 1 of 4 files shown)


hemberger:
Also, here's what the Twitter iframe looks like on Firefox, where we don't get the scrollbar styling:
image
It's not great, but it's still completely functional.

It's also worth noting that the browsers where we don't get the cool styling account for ~7.5% of the browser marketshare (Firefix at 5%, and IE at 2.5%).

PR #653, opened 2019-03-03
Scrutinizer fixes

hemberger:
Add more fixes of Scrutinizer CI alerts.


lib/Default/AbstractSmrShip.class.inc
```diff
@@ -145,13 +145,12 @@ abstract class AbstractSmrShip {
...
-			while(($total=array_sum($goods=$this->getCargo())) > $this->getCargoHolds()) {
-				do {
-					list($goodID,$amount)=each($goods);
-				} while($amount==0);
-				if($amount > $total-$this->getCargoHolds())
-					$amount = $total-$this->getCargoHolds();
-				$this->decreaseCargo($goodID,$amount);
+			foreach ($this->getCargo() as $goodID => $amount) {
+				$excess = array_sum($this->getCargo()) - $this->getCargoHolds();
+				if ($excess > 0) {
```
Page-: It would be good to add a
```php
} else {
	break; // or return;
}
```
so that you stop the iteration as soon as there is no excess.  You could also avoid the repeated array_sum calls with `$excess -= $decreaseAmount`

hemberger: Good idea, thanks!

PR #658, opened 2019-03-08
Create a new Template for each page

hemberger:
If we initialize the Template at the config.inc level, then it will
be reused between pages that call `forward`. In general, this is not
a problem, because we typically don't set template variables until
we're on the final page that we intend to display. However, it has
caused a problem in at least one case, where we forwarded to the
error.php page after the `PageTopic` variable had already been set
(thus triggering an Exception).

This particular issue could have been obviated by not setting the
`PageTopic` template variable until after all `create_error` sanity
checks had been performed on the display page. But this is not a
safe assumption (it depends on the coder always remembering when to
assign to the template), nor does it fix any cases that we haven't
yet encountered.

Instead, creating the template in `do_voodoo` ensures that we have
a fresh template for each page, guaranteed to be free of any stale
template variables.

(This does require that we also create a Template on any page that
doesn't use `do_voodoo`, e.g., the Galaxy Map and the login page.)

### DIFF
--- a/htdocs/config.inc
+++ b/htdocs/config.inc
@@ -120,8 +120,3 @@ require_once('config.php');
 require_once(ROOT . 'vendor/autoload.php');
 require_once(LIB . 'autoload.inc');
 spl_autoload_register('get_class_loc');
-
-// Initialize the template
-$template = new Template();
-$GLOBALS['template'] =& $template;
-
--- a/htdocs/login.php
+++ b/htdocs/login.php
@@ -26,6 +26,7 @@ try {
 		exit;
 	}
 	
+	$template = new Template();
 	if(isset($_REQUEST['msg']))
 		$template->assign('Message',htmlentities(trim($_REQUEST['msg']),ENT_COMPAT,'utf-8'));
 
--- a/htdocs/map_galaxy.php
+++ b/htdocs/map_galaxy.php
@@ -64,6 +64,9 @@ try {
 		throw new Exception('Failed to start session');
 	}
 
+	// Initialize the template
+	$template = new Template();
+
 	// Set temporary options
 	if ($player->hasAlliance()) {
 		if (isset($_POST['change_settings'])) {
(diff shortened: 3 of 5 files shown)


hemberger:
I like this idea, especially because there may be places where we were expecting to forward to a display page with a template variable assigned, but where we check for existence of the variable before displaying it, and thus would _**silently**_ omit it -- e.g. `if (isset($SomeForwardedVar)) { ... }`. (For example, 846a77a9d7f77c44b2fc4d1cd4edb49e218626af indicates such practices might be in use somewhere.)

I've got a simple implementation in mind:
```php
    public function __destruct() {
        if (!$this->displayCalled && !empty($this->data)) {
            error_log('Template destroyed before displaying the following assigned keys: ' . join(', ', array_keys($this->data)));
        }   
    }   
```
So we'd see a message that looks like:
> smr-game     | [Fri Mar 08 16:44:13.380446 2019] [php7:notice] [pid 16] Template destroyed before displaying the following assigned keys: PageTopic, MenuBar, referer: http://localhost/loader.php?sn=cc5430f1

The problem is that this doesn't track where these variables were set, which makes the warning much less useful in cases where the variable names are generic (like in the above example). I'm not sure if it's possible to add such bookkeeping in a sensible way -- without modifying every call to `$template->assign(...)`. Do you have any ideas?

Page-:
You can either do something like
```php
	private $assigners = array();

	public function assign($var, $value) {
		if (!isset($this->data[$var])) {
			$this->data[$var] = $value;
			$this->assigners[$var] = new Exception();
		} else {
			echo $this->assigners[$var];
			// We insist that template variables not change once they are set
			throw new Exception("Cannot re-assign template variable '$var'!");
		}
	}
```
or
```php
	private $assigners = array();

	public function assign($var, $value) {
		if (!isset($this->data[$var])) {
			$this->data[$var] = $value;
			$this->assigners[$var] = debug_backtrace();
		} else {
			echo $this->assigners[$var];
			// We insist that template variables not change once they are set
			throw new Exception("Cannot re-assign template variable '$var'!");
		}
	}
```
in order to capture a stack trace at the point of assigning so that you can print it when either reassigning or destroying.  I don't know which of the two methods is more performant but regardless of which is you almost certainly want to hide it behind a debug flag as I would expect both to introduce a noticeable performance penalty - but it could save a lot of time whilst debugging

PR #743, opened 2019-06-07
filter_list.js cleanup

Page-:
* Remove unnecessary `in_array` function
* Avoid unnecessarily referencing `window`
* Avoid the need to prepoulate the filter array

These changes should make the script a little easier to work with and also avoids modifying the array prototype (generally something to avoid), I haven't tested these because live SMR still has the old html so I can't drop in my changes to test but can do so once it's deployed if you want to wait

### DIFF
--- a/htdocs/js/filter_list.js
+++ b/htdocs/js/filter_list.js
@@ -1,15 +1,5 @@
 // For use with {ship,weapon}_list.php
 
-// Benoit Asselin - http://www.ab-d.fr
-Array.prototype.in_array = function(p_val) {
-	for(var i = 0, l = this.length; i < l; i++) {
-		if(this[i] == p_val) {
-			return true;
-		}
-	}
-	return false;
-}
-
 // Use window variable to store filter values. Since this is used for both
 // the ship and weapon tables, it needs to have enough elements for either.
 window.filter = ["All", "All", "All", "All", "All", "All", "All", "All", "All", "All",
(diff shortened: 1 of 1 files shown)


Page-:
Changing `== null` will actually most likely break the code - I use it to check for either null or undefined here. It's also generally recognised as the only exception too using triple equals as standard in javascript

hemberger:
Indeed, `===` does break your patch, thanks for catching that.

Would you object to an explicit `=== undefined || === null` instead? I recognize that it's more verbose, but `== null` is a subtle feature that at least warrants a comment for people like me (_and_ I have an unhealthy obsession with fixing Scrutinizer issues!).

Page-:
`== null` is a pretty standard way to check nullish in javascript and is usually accepted by linters with an option to be extra strict because it's a common way of checking so I'm guessing you've just configured scrutinizer to be overly strict (`eslint:recommended` is probably a good set of rules to go with)

The only alternative I see being used at all is `!x` which actually checks for `x == null || x === '' || x === 0 || ...` but can be equivalent if you know the other options aren't possible (or if they should also be handled the same way).

Basically I think it's better to allow `== null` and treat it as a language feature that is part of using the language for better or worse (like `!!x` in C and some other languages for converting to a boolean) but if you still feel strongly against it I can change it

hemberger:
Unfortunately I don't know how to modify the linting options in Scrutinizer, so I think we'll just have to live with the warning for now, which is fine. (Btw, I had considered using `!x` instead, but we want to explicitly filter on the empty string as well.)

Thanks again for the PR!

PR #751, opened 2019-06-21
Add an Admin Tool to manage NPCs

hemberger:
This admin tool replaces some of the logic in the `npc.php` script to create NPC accounts/players/alliances. It also allows us to add new NPC slots and to set their active status, which previously could only be done directly inside the database. The name and alliance specified in `npc_logins` are now only default values -- these can be changed at creation time (so that, e.g., you don't need to create an entirely new NPC account just to have a different NPC name).

Some engine pages needed to be refactored into class methods (`SmrAlliance::createAlliance` and `SmrPlayer::createPlayer`) and several other methods needed to be added.

See commit messages for complete details.

image

### DIFF
--- a/admin/Default/npc_manage.php
+++ b/admin/Default/npc_manage.php
@@ -1,0 +1,60 @@
+<?php
+
+$template->assign('PageTopic', 'Manage NPCs');
+
+$selectedGameID = SmrSession::getRequestVar('selected_game_id');
+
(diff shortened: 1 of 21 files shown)


hemberger:
This PR now also includes some basic fixes to the `npc.php` script so that it is more-or-less functional.

PR #765, opened 2019-09-22
Update shop_goods_processing.php

stupidnewbie:
Moved relation gain under buy and sell conditionals so that stealing no longer results in relation gains


Page-:
That's a hilarious bug!  Could potentially also remove relations on stealing? (it'd be thematic for roleplay, but I expect unlikely to have any real gameplay effect)

hemberger:
I agree! It would be nice to decrease relations if you get caught stealing (currently it only decreases alignment). If you're not caught, then... what they don't know can't hurt you. ;)

PR #768, opened 2019-09-29
SmrPort: stolen goods bugfix

stupidnewbie:
Stolen goods has been working towards a port level increase.  Added a separate function for stealing goods to the port class so that theft bypasses port upgrade calls.

### DIFF
--- a/engine/Default/shop_goods_processing.php
+++ b/engine/Default/shop_goods_processing.php
@@ -126,7 +126,7 @@ if ($transaction == 'Steal' ||
 		$ship->increaseCargo($good_id, $amount);
 		$player->increaseHOF($amount, array('Trade', 'Goods', 'Stolen'), HOF_ALLIANCE);
 		$player->increaseHOF($gained_exp, array('Trade', 'Experience', 'Stealing'), HOF_PUBLIC);
-		$port->buyGoods($portGood, $amount, $ideal_price, $bargain_price, $gained_exp);
+		$port->stealGoods($portGood, $amount);
 	}
 
 	$player->increaseHOF($gained_exp, array('Trade', 'Experience', 'Total'), HOF_PUBLIC);
--- a/lib/Default/AbstractSmrPort.class.inc
+++ b/lib/Default/AbstractSmrPort.class.inc
@@ -397,7 +397,7 @@ class AbstractSmrPort {
 	}
 	
 	protected function stealGoods(array $good, $goodsTraded) {
-		$this->decreaseGood($good, $goodsTraded, true);
+		$this->decreaseGood($good, $goodsTraded, false);
 	}
 	
 	public function checkForUpgrade() {
--- a/lib/Default/AbstractSmrPort.class.inc
+++ b/lib/Default/AbstractSmrPort.class.inc
@@ -396,6 +396,10 @@ class AbstractSmrPort {
 		$this->tradeGoods($good, $goodsTraded, $exp);
 	}
 	
+	protected function stealGoods(array $good, $goodsTraded) {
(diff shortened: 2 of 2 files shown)


hemberger:
Thanks for the PR! This looks good to me, but it also makes me wonder if we should use `decreaseGoodAmount` (or `decreaseGood(..., false)`) to skip the pumping of higher level goods. What do you think @stupidnewbie?

stupidnewbie:
I see no reason for theft to increase supply of more expensive goods.  I would argue it's more intuitive for theft to hamper supply.  I'll make the change if you agree. 

From a gameplay point of view recent PRs have been a serious blow to stealing.  I think we are going from an overpowered mechanism to something nobody will use.  Perhaps consider allowing evil traders to steal regardless of ship?  And add a risk reduction mechanic to the evil ships similar to smuggled goods?

hemberger:
If you're on board, I'd say go ahead and make the appropriate change to `stealGoods` so that it no longer increases supply of higher level goods.

As for whether or not these changes make stealing unusable, I am uncertain. Obviously it will no longer be a mechanism to efficiently increase relations, but you still get perfect xp and avoid buying costs. If anything, I'd probably prefer simply making the Thief a slightly better trader. What scenarios are you most worried about losing?

stupidnewbie:
Simply buying the goods is only prohibitively expensive when relations are low.  In which case the relations need to be improved.  Improving relations makes a more meaningful difference in profit and provides a strong incentive to buy.

Increasing port levels is important early-mid game in order to get access to higher tier goods.  Buying is more effective than selling and stealing will not help at all.

As the game goes on trade ships become larger and faster.  Resupply of higher tier goods is an effective way to keep your route supplied so there is a route to spend our turns on.  Buying may cost you 5-10% of your profit over stealing but you're going to get some 20% more runs across the route before it's dry.

I'll experiment with it and maybe I'm missing something.  But I see that trio of changes making stealing a circumstantial choice at best.

I would argue the thief should be given a trading buff considering the cost of it both in terms of cash and alignment.  But I would first ask how ships are balanced now?  Are usage metrics gathered?  Is there a weighted point system for stats?  Do we just kind of wing it?

hemberger:
I agree completely that it makes smuggling ships a more circumstantial choice. However, if you're in a smuggling ship, it's likely that the only times you won't steal when you would have previously are when you are specifically trying to a) increase relations or b) pump higher level goods. It continues to improve the value of trading T3 goods for profit. It also makes it so that stealing isn't always the right choice even if you have the option, which I think is a good thing.

With the current mechanics of port leveling, stealing (along with bargaining at the ideal price) are slow to level ports. Instead, you would likely overspend when buying goods, since there is an upper limit of `$numGoods * 1000` credits for the upgrade, which is almost always well above the ideal price. Though in some cases, you might have been able to advance a level 7 or 8 port by stealing if it was selling a T3 good with a very large distance multiplier.

It's also important to keep in mind that stealing has not been around long. I added it in March 2018, and only a handful of people have really taken advantage of it since then (which you can see from the Hall of Fame). So I would not consider it a sacred institution that needs to be preserved.

Your points are all well-taken, but at the end of the day, these were unintentional oversights, and I think you have rightly labeled them as "bugs". Let's fix the bugs, and then we can focus on finding a way for alternative trading styles be competitive. Does that seem reasonable?

------

Balancing ships is beyond the scope of this issue, so let's discuss it elsewhere, but I will just quickly say that I have been extremely conservative during my tenure as maintainer, making only a few minor ship balancing changes. It doesn't _need_ to stay that way though.

stupidnewbie:
Agreed on bugs vs balancing.  This is an unintentional behavior repaired.

PR #769, opened 2019-10-04
Use the .php extension for PHP class files

hemberger:
This is largely a matter of preference, but by using .php (instead of
.inc) there is no additional configuration that is needed for editors
to use the correct syntax highlighting. The .inc extension is less
popular in the modern era of PHP.


hemberger:
Scrutinizer reports new issues, but they are not new. It just failed to detect the filename changes.

PR #775, opened 2019-10-10
Prevent overwriting planet mounted weapons

hemberger:
When adding a mounted weapon, it is possible that a weapon has been
added to the selected slot in the time between when the "Defense" page
was rendered and when the "Transfer" button was clicked.

To mitigate this (rare) scenario, the processing script will now raise
an error if the selected slot is already occupied.

Also refactor the common logic from moveMountedWeapon{Up,Down} into a
new function swapMountedWeapons.


hemberger:
Note that the "Defense" page will not ajax-update when there are input boxes (e.g. any gens/bunkers/hangars), which makes it more likely for this issue to be encountered.

PR #783, opened 2019-11-02
weapon hit code cleaning

stupidnewbie:
I tweaked the weighted random code slightly.  I believe weapons will hit at their intended rate regardless of their accuracy with this change.

I also changed the code to state things in terms of affirmation.  A hit is true and a miss is false.  Hopefully this is more direct and easier to read.

### DIFF
--- a/lib/Default/SmrWeapon.class.php
+++ b/lib/Default/SmrWeapon.class.php
@@ -103,9 +103,12 @@ class SmrWeapon extends AbstractSmrCombatWeapon {
 	protected function &getWeightedRandomForPlayer(AbstractSmrPlayer $player) {
 		return WeightedRandom::getWeightedRandomForPlayer($player, 'Weapon', $this->getWeaponTypeID());
 	}
-	
-	protected function checkMiss(AbstractSmrPlayer $player, $percent) {
-		return $this->getWeightedRandomForPlayer($player)->percentFailed($percent);
+
+	/**
+	 * Given $weaponAccuracy as a percent, decide if the weapon hits.
+	 */
+	protected function checkHit(AbstractSmrPlayer $player, $weaponAccuracy) : bool {
+		return $this->getWeightedRandomForPlayer($player)->flipWeightedCoin($weaponAccuracy);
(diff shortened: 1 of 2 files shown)


lib/Default/WeightedRandom.class.inc
```diff
@@ -73,19 +73,15 @@ class WeightedRandom {
...
-		$failed = mt_rand(0, 99) + $this->weighting >= $percent;
-		$this->adjustWeightByPercent(!$failed, $percent);
-		return $failed;
-	}
-	
-	public function adjustWeightByPercent($success, $percent) {
-		$percent = $percent / 100;
-		if ($success) {
-			$percent = -1 + $percent;
+	public function checkHit($weaponAccuracy) {
```
Page-: This code looks equivalent to before but just inverted to me? Am I missing something as you mentioned you tweaked it

stupidnewbie: I believe you are correct.  Looking back I suspect my testing was at fault for the numbers I was getting.

lib/Default/WeightedRandom.class.inc
```diff
@@ -73,19 +73,14 @@ class WeightedRandom {
...
```
hemberger: This class was previously context-agnostic. Is it possible to rename the function and variables to maintain that? (Perhaps something like `checkSuccess`, `$percentChance`, and `$success` or whatever seems suitable.)

lib/Default/SmrWeapon.class.inc
```diff
@@ -249,44 +245,52 @@ class SmrWeapon extends AbstractSmrCombatWeapon {
 		return $damage;
 	}
 	
+	public function checkHit(AbstractSmrPlayer $player, $weaponAccuracy) {
+		$this->getWeightedRandomForPlayer($player)->flipWeightedCoin($weaponAccuracy);
```
hemberger: This is missing a return value. If you used this code in testing the changes, then you probably should have seen misses 100% of the time.

PR #784, opened 2019-11-08
add trap to catch accuracy 100 or greater

stupidnewbie:
when firing at 100 or greater accuracy there is no need for a weighted random roll.  the shot is already guaranteed to hit.  simply reporting the hit saves resources and should prevent the weight from being adjusted the wrong way.


Page-:
Ah, so if the weapon is >100% accurate it actually becomes more likely to hit with each shot, which you could store up and whilst that's not usually an issue it could potentially cause you too be too accurate after dying and losing levels such that you become <100% accurate. It's pretty niche but worth fixing

PR #800, opened 2019-12-12
npc.php: add NPC_MAX_ACTIONS

hemberger:
To avoid an infinite loop, set an upper limit for the number of
actions an NPC can take in a single execution.

The sample limit is set to 2500 actions, which corresponds to about
a half hour of actions.

### DIFF
--- a/config/npc/config.specific.sample.php
+++ b/config/npc/config.specific.sample.php
@@ -2,6 +2,7 @@
 
 const NPC_LOG_TO_DATABASE = true; // insert debug messages into db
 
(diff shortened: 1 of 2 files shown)


hemberger:
For reference, without the max actions, an update to the live engine between games caused the npc's to get stuck in a loop between reshipping and plotting to Fed HQ (which I have yet to be able to reproduce). This caused the crons to pile up and create ~GB log files. Fortunately this only occurred for a few hours before the previous game ended.

PR #804, opened 2019-12-17
Weapon power handling change

stupidnewbie:


stupidnewbie:
a devastator, with 26 power, should be able to mount 8 level 3 weapons for a total of 24 power.  current code checking does not permit that.  it treats the first slot as up to level 5, next two slots as up to 4, next 3 up to 3, and so on.

either the concept of ship power is misleading or this is unintended behavior.  probably easier to sort it out this way.

lib/Default/SmrShip.class.php
```diff
@@ -153,30 +153,11 @@ public function update() {
...
-			break;
-			case 2:
-				return $this->db->getNumRows() < 4;
-			break;
-			case 1:
-				return true;
-			break;
-		}
-		return true;
+	/*
```
hemberger: The formatting for this docstring is slightly off. It should look like:
```
/**
 * Two asterisks on the first line (with one it's just a comment, not a docstring).
 * Text line asterisks aligned with the first asterisk on the first line.
 * One space between text-line asterisks and the text.
 */
```

lib/Default/SmrShip.class.php
```diff
@@ -153,30 +153,11 @@ public function update() {
...
+	*	given power level of new weapon, return whether there is enough power available to install it on this ship
+	*/
+	public function checkPowerAvailable($powerLevel) {
+		return this->getRemainingPower() >= $powerLevel;
```
hemberger: I think Scrutinizer is complaining here because `this` needs to be `$this`.

hemberger:
Oh, and doesn't this also imply that you could have, for example, 5 Level 5 weapons + 1 Level 1 weapon on a 26 power ship? For better or worse, this would make re-stocking weapons on Protoplanets easier...

stupidnewbie:
I believe I've got those two things cleared up.  And I took a stab at squashing my commits but I'm not sure if it worked. 

Upon further inspection, this will add a fair bit of depth to weapon selection.  It will be a balancing act between hardpoint efficient high level weapons and power efficient low level weapons.  There will also be a greater number of working combinations which is good news for those with limited racial options.  The additional options will make the choice more meaningful which should make a fun game mechanic.

lib/Default/SmrShip.class.php
hemberger: On the updated version, you got the first item there (two asterisks), but the spacing is off (asterisks aren't aligned and you have a tab instead of a space between the asterisk and text).

PR #829, opened 2020-02-17
Alliance account

stupidnewbie:
Display the amount of credits in alliance account on trader status screen.  Because my spaceship should have that kind of capacity for communication. 

Assuming it works of course.  I have no local test bench.


hemberger:
This won't work if the player is not in an alliance. You will want to make it conditional on `$ThisPlayer->hasAlliance()`.

The original reason for not having the alliance bank displayed anywhere in the universe was to make it more dangerous to get this information (since hunters often watched banks). However, now that this information is available in Discord, perhaps this logic no longer holds.

This is a gameplay change, so it will need to be approved by an admin and/or the community.

PR #830, opened 2020-02-17
channel_msg_money.php: use SmrAlliance methods

hemberger:
Get the total credits in the alliance bank account by using the
`SmrAlliance::getAccount` method instead of a raw SQL query.

Also document the method since its name is a bit ambiguous.

### DIFF
--- a/lib/Default/SmrAlliance.class.php
+++ b/lib/Default/SmrAlliance.class.php
@@ -244,6 +244,9 @@ class SmrAlliance {
 		$this->imgSrc = htmlspecialchars($url);
 	}
 
+	/**
+	 * Get the total credits in the alliance bank account.
+	 */
 	public function getAccount() {
 		return $this->account;
 	}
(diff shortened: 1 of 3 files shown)


lib/Default/SmrAlliance.class.php
```diff
@@ -244,6 +244,9 @@ public function setImageURL($url) {
 		$this->imgSrc = htmlspecialchars($url);
 	}
 
+	/**
+	 * Get the total credits in the alliance bank account.
```
Page-: It might make more sense to just rename the function so it's clear wherever it's called without even needing to look at the docs, eg getAccountBalance?

hemberger: It definitely would! I just didn't want to make more work for the author of #829 (who is, surprisingly, not me).

stupidnewbie: For the record I don't mind doing extra work in the name of making things orderly and sensible.

hemberger: Then your wish is my command! I've renamed `SmrAlliance::getAccount` to `SmrAlliance::getBank` (for consistency with SmrPlayer methods).

PR #839, opened 2020-04-02
Uni Gen: add dropdown to quickly switch galaxies

hemberger:
In the "Create {Locations,Planets,Ports}" sub-pages, instead of just
saying what galaxy we're working on, turn it into a dropdown where you
can quickly navigate to other galaxies.


hemberger:
image

PR #842, opened 2020-04-06
Add open and closed recruiting options for alliances

hemberger:
When creating and modifying alliances, there will now be three
recruitment options:

1) Open - Anyone can join freely, no password is needed
2) Closed - Players can only join by direct invitation
3) Password - Players can join by invitation or with a password

This creates a lot more flexibility for alliance leaders to manage
their membership.

Note: We use a combination of the `password` and `recruiting`
attributes of SmrAlliance to infer the "recruitment type". This
is perhaps not ideal, but it requires no database changes.


hemberger:
Change Alliance Stats
image

image

Create Alliance
image

Alliance Roster
image

image

image

PR #866, opened 2020-06-14
Update SmrWeapon.class.php

stupidnewbie:
ignore maneuverability rating difference values outside of what standard ships will generate.  AKA ignore the SMS for this bit of code.

standard ship manuverability rating ranges around 550-750.  the midnight special MR is -8920 which is enough to force weapon accuracy in to negative values during ship to ship combat.


lib/Default/SmrWeapon.class.php
```diff
@@ -153,7 +153,8 @@ public function getModifiedAccuracyAgainstPlayer(AbstractSmrPlayer $weaponPlayer
 		$weaponShip = $weaponPlayer->getShip();
 		$targetShip = $targetPlayer->getShip();
 		$mrDiff = $targetShip->getMR() - $weaponShip->getMR();
-		if ($mrDiff > 0) {
+		//the 500 cap is to exclude huge numbers generated by the SMS which can cause accuracy to become absolute
```
Page-, line 156: I think it'd make more sense to do
```php
$mrDiff = min($mrDiff, 500); // Cap the mr diff to 500
```

stupidnewbie: That is a scalpel to my chain saw.  I like it.

That will result in a 500MR difference instead of ignoring MR entirely.  A 500 MR difference will result in accuracy being degraded by 33% when a SMS fires upon a standard ship.

At second glance, I believe it would be better if negative MR values weren't returned in the first place.  I'm going to update this accordingly.

lib/Default/AbstractSmrShip.class.php
```diff
@@ -1089,8 +1089,13 @@ protected function doArmourDamage($damage) {
...
 	}
 
+	/**
+	 * Returns the maneuverability rating for this ship.
+	 */
 	public function getMR() : int {
-		return IRound(
+		//700 - [ (ship hit points / 25) + (ship stat factors) ]
+		//Minimum value of 0 because negative values cause issues with calculations calling this routine
+		return Max(0, IRound(
```
hemberger, line 1098:
```suggestion
		return max(0, IRound(
```
PHP, for who knows what reason, is case-insensitive. However, I prefer to use the "canonical" capitalization, which in this case is lowercase.

PR #867, opened 2020-06-16
Balance L4 weapon power efficiency

stupidnewbie:
L4 weapons are 5-10% less power efficient than L5 weapons. This is out of step with the other power level changes.

L3 weapons are around 20% more power efficient than L5 weapons. L2 weapons are around 70% more power efficient as compared to L5. and L1 around 120% more efficient than L5.

This change will move L4 weapons in to the range of 5-10% more power efficient than L5.

### DIFF
--- a/db/patches/Vunknown__update_level4_weapons.sql
+++ b/db/patches/Vunknown__update_level4_weapons.sql
@@ -1,0 +1,10 @@
+-- Increase damage of Level 4 weapons by 15% to bring power efficiency in step with other power levels
+UPDATE weapon_type SET armour_damage = 230 WHERE weapon_name = 'Big Momma Torpedo Launcher';
+UPDATE weapon_type SET armour_damage = 115 WHERE weapon_name = 'Little Junior Torpedo';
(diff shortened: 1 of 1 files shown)


hemberger:
Hey @stupidnewbie thank for the PR! I agree that there is still an inconsistency in the power efficiency. At this time, however, I think I'd like to leave the weapons where they are. This is for a couple reasons:
1. I'd like to get a bit more usage data, so I don't want to make more changes before we have a chance to see the impact of the L1 weapon changes.
2. I would prefer to decrease damage profiles overall, rather than increase them. I think a more likely scenario is that L5 weapon accuracies would decrease somewhat.

I'm still not entirely sure how power efficiency should scale with weapon level. Should it be constant, or should it increase with decreasing level?

Also, I've been thinking about a different mechanic entirely, where L5 weapons would be ultra-low accuracy, but there would be a huge +% accuracy bonus when attacking stationary objects (ports/planets). This would make "bunker buster" L5 weapons something you only use for port raids and planet busts, and you would need a different loadout for ship vs. ship combat.

PR #869, opened 2020-06-28
Changes for Draft Rounds

hemberger:
* Draft leaders may be associated with "home sectors" when assigned, which is where players will be moved to (if in Sector 1 to start), and where they will return when podded.
* Players are eligible to be picked if they are in Sector 1, so long as they have not previously been drafted.

Some minor refactoring of `SmrPlayer` was needed to achieve this behavior.


hemberger:
image

PR #871, opened 2020-06-28
Add ability to change player race

hemberger:
Players will be able to change their race once if done within the
first 48 hours of the game. This is done via Preferences.

Presently, when changing races, you will be:
* moved to your new home sector (Racial HQ, etc.)
* reset to starting alignment
* reset to starting ship
* given 1 newbie turn if you have none

Explored sectors, kills, credits, and HOF are not changed (since
these things can't easily be undone or can be easily circumvented,
e.g. sharing maps, alliance/anon bank accounts, etc.).

Also needed to add a few functions to avoid code duplication.

image


hemberger:
This also creates a news entry:
image

PR #875, opened 2020-07-19
Handle overflow consistently in bank transactions

hemberger:
Modify credit-related transaction functions so that they don't
overflow the max value for the database. They now return the actual
amount added (which may be reduced from the original value).

* SmrAlliance::increaseBank (new, along with `decreaseBank`)
* SmrPlayer::increaseCredits
* SmrPlayer::increaseBank
* SmrPlanet::increaseCredits

Adds informational message "(Account is Full)" to the Personal Bank
and Planet Financial pages to indicate when the max credit amount has
been reached.

Note: Anonymous Accounts can still overflow, because they modify the
database directly and don't have a convenient class interface.


hemberger:
image

PR #876, opened 2020-07-20
Tweaks to message boxes

hemberger:

Scout message UI/UX changes:

* The sender will no longer be displayed as a separate (redundant) field so
  that the message can fit on a single line.

Message box UI/UX changes:

* The "Political Messages" box now always shows up in box view.
  Conditionally hiding it seemed unnecessary.

* Remove redundant link to "Manage Blacklist" in box view. We keep the
  link in the navigation bar.

* Clicking "Empty Read Messages" for Scout messages used to delete all
  Scout messages (even unread ones). Now it behaves like all other box
  types, and only deletes read messages.

* Deleting messages now stays in the selected folder instead of going
  back to the box view, unless deleting "All Messages" (since it does
  not make sense to display an empty list of messages).

* Reporting a message now returns to the currently selected folder.

### DIFF
--- a/engine/Default/message_box.php
+++ b/engine/Default/message_box.php
@@ -1,0 +1,53 @@
+<?php declare(strict_types=1);
+
+require_once(get_file_loc('message.functions.inc'));
+Menu::messages();
+
+$template->assign('PageTopic', 'View Messages');
+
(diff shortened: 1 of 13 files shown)


hemberger:
image

PR #882, opened 2020-08-23
Update AbstractSmrPort.class.php

stupidnewbie:
Gain 15% of the difference between the amount of XP normally earned and the whole amount.

This will give traders a portion of XP even at low relations.  Intended effect is to lessen frustration related to bartering or grinding relations.

15% is a wild ass guess.  May need some tweaking.

### DIFF
--- a/engine/Default/shop_goods_processing.php
+++ b/engine/Default/shop_goods_processing.php
@@ -120,7 +120,7 @@ if ($transaction == 'Steal' ||
 	$base_xp = SmrPort::getBaseExperience($amount, $port->getGoodDistance($good_id));
 
 	// if offered equals ideal we get a problem (division by zero)
-	$gained_exp = IRound($port->calculateExperiencePercent($ideal_price, $offered_price, $bargain_price, $transaction) * $base_xp);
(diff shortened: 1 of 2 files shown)


stupidnewbie:
I took a stab at the message.  Here's to untested code!

I had originally thought 5 or 10% would do it.  But then I realized for most trades, even in a late game trader, this would be single digit XP gains.  Which feels bad imo.  15% should dip in to double digits here and there with 300 holds and the usual mediocre routes.  And even at 15% someone at high relations will still gain way more.

That said it may need tweaking.  Or refactoring.  Or to have a stake driven through it.  Time shall tell.

hemberger:
@stupidnewbie Instead of adding yet another modifier onto an already overly-complicated formula, I've taken this opportunity to simplify all the bargaining formulas. Everything is now linear, which means that, for example, at 500 relations, you get half the experience of 1000 relations. I feel that's pretty generous (especially compared to the `(ideal - bargain)^13` factor!). If you have no objections to this change, I'll push it to beta for further testing.

stupidnewbie:
I believe any change that makes bartering less user hostile is a boon.  Allowing for a sidestep plus making the process simpler will absolutely help.

PR #886, opened 2020-09-02
Cloak on examine

stupidnewbie:
add a unique message indicating a ship being examined has cloaked.  in contrast, the current behavior is the catch all "You have no targets!" message which may be confusing.


templates/Default/engine/Default/trader_examine.php
```diff
@@ -15,6 +15,8 @@
 	?><p class="big blue">You are under federal protection! That wouldn't be fair.</p><?php
 } elseif ($TargetPlayer->hasFederalProtection()) {
 	?><p class="big blue">Your target is under federal protection!</p><?php
+}elseif (!$ThisPlayer->canSee($TargetPlayer)) {
```
hemberger, line 18:
```suggestion
} elseif (!$ThisPlayer->canSee($TargetPlayer)) {
```

hemberger (approved):

PR #887, opened 2020-09-03
135 balance change proposal

stupidnewbie:
I'm throwing this out there for whatever use or non use.  My aim is to cut down the profits from port raids, encourage trading, balance out P5 weapons a bit, and to slow de-mining somewhat.  Especially from fed ships.

Holy Hand Grenade and Nuke damage reduced by 10% from 300 to 270.  This brings damage per shot with stacked P5 guns down to numbers similar to what they were before the power slot change.  I am targeting the damage and not accuracy because the swing shot/burst damage needs to come down as well for lower power weapons to be decent substitutes.

Port payouts down 25%.  This method of farming cash is driving too much in to the game economy and too quickly.

Trade profits boosted up to +15% from pre 134.  To further encourage trading for money.

Mine armor increase from 20 to 25.  This takes a full force stack from 1250 armor to 1500 armor.  So 1-3 more weapon strikes per stack.  This should make it harder to bump + one shot stacks.

Fed ship turn bonus against forces removed.  Fed ships are already very capable of ramming down forces with a damage reduction.

### DIFF
--- a/db/patches/V1_6_64_06__update_lvl5_weapons.sql
+++ b/db/patches/V1_6_64_06__update_lvl5_weapons.sql
@@ -1,0 +1,3 @@
+-- Decrease damage of HHG/Nuke by 10%
+UPDATE weapon_type SET armour_damage = 270 WHERE weapon_name = 'Nuke';
+UPDATE weapon_type SET shield_damage = 270 WHERE weapon_name = 'Holy Hand Grenade';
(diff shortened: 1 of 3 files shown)


stupidnewbie:
I've dropped the mine based commits.  Everything past a week in to the game is driven by massive port raid profits.  Reducing that may have a substantial effect on forces by its self.

PR #895, opened 2020-09-24
add additional bounty for newbie kills

stupidnewbie:
added bounty of 200 times experience for killing poorly armed newbies (read trade ships).  no stats are awarded as part of this additional bounty as to not encourage notoriety seeking.  adjusted news feed to notify others of newbie kill and the bounty that comes along with it.

there's not much immediate need for this but it's an accepted feature request and I need the practice coding.

### DIFF
--- a/lib/Default/AbstractSmrPlayer.class.php
+++ b/lib/Default/AbstractSmrPlayer.class.php
@@ -544,6 +544,17 @@ abstract class AbstractSmrPlayer {
 	public function hasNewbieStatus() {
 		return $this->newbieStatus;
 	}
+	
+	/**
+	 * Does the player have newbie status and
+	 * a low attack rating?
+	 */
+	public function isNoncombatNewbie() {
+		if ($this->getShip()->getAttackRatingWithMaxCDs() <= MAX_ATTACK_RATING_NEWBIE && $this->hasNewbieStatus()) {
+			return true;
+		}
+		return false;
+	}
 
 	/**
 	 * Update the player's newbie status if it has changed.
@@ -2216,6 +2227,9 @@ abstract class AbstractSmrPlayer {
 		$this->setNewbieTurns(100);
 	}
 
+	/*
+	 * process $this player being podded by $killer
+	 */
 	public function &killPlayerByPlayer(AbstractSmrPlayer $killer) {
 		$return = array();
 		$msg = $this->getBBLink();
@@ -2225,16 +2239,16 @@ abstract class AbstractSmrPlayer {
 			$msg .= ' flying <span class="yellow">' . $named_ship . '</span>';
 		}
 		//adjust message for newbie kill
-		if ($this->getShip()->getAttackRatingWithMaxCDs() <= MAX_ATTACK_RATING_NEWBIE && $this->hasNewbieStatus() && !$killer->hasNewbieStatus()) {
+		if ($this->isNoncombatNewbie() && !$killer->hasNewbieStatus()) {
 			$msg .= ' was newbie killed by ' . $killer->getBBLink();
 		} else {
 			$msg .= ' was destroyed by ' . $killer->getBBLink();
+			if ($killer->hasCustomShipName()) {
+				$named_ship = strip_tags($killer->getCustomShipName(), '<font><span><img>');
+				$msg .= ' flying <span class="yellow">' . $named_ship . '</span>';
+			}
 		}
 		
-		if ($killer->hasCustomShipName()) {
-			$named_ship = strip_tags($killer->getCustomShipName(), '<font><span><img>');
-			$msg .= ' flying <span class="yellow">' . $named_ship . '</span>';
-		}
 		$msg .= ' in Sector&nbsp;' . Globals::getSectorBBLink($this->getSectorID());
 		$this->getSector()->increaseBattles(1);
 		$this->db->query('INSERT INTO news (game_id,time,news_message,type,killer_id,killer_alliance,dead_id,dead_alliance) VALUES (' . $this->db->escapeNumber($this->getGameID()) . ',' . $this->db->escapeNumber(TIME) . ',' . $this->db->escapeString($msg, true) . ',\'regular\',' . $this->db->escapeNumber($killer->getAccountID()) . ',' . $this->db->escapeNumber($killer->getAllianceID()) . ',' . $this->db->escapeNumber($this->getAccountID()) . ',' . $this->db->escapeNumber($this->getAllianceID()) . ')');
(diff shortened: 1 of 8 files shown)


hemberger (changes requested):
I like the spirit of this change. Thanks for taking the initiative!

A few suggestions:
1. Add a note under the attack button on the Examine screen warning the attacker that this is a "newbie kill".
2. Refactor the condition for deciding if it's a newbie kill into a method on `SmrPlayer` so that we don't duplicate it all over the place.

Also, do we want to penalize the killer further? No exp from kill? No credits from kill? It already doesn't count toward kill rankings (or HoF).

lib/Default/AbstractSmrPlayer.class.php
```diff
@@ -542,6 +544,17 @@ public function isNPC() {
 	public function hasNewbieStatus() {
 		return $this->newbieStatus;
 	}
+	
+	/**
+	 * Does the player have newbie status and
```
hemberger, line 549: This comment can fit on a single line.

lib/Default/AbstractSmrPlayer.class.php
```diff
@@ -542,6 +544,17 @@ public function isNPC() {
...
+	 * a low attack rating?
+	 */
+	public function isNoncombatNewbie() {
+		if ($this->getShip()->getAttackRatingWithMaxCDs() <= MAX_ATTACK_RATING_NEWBIE && $this->hasNewbieStatus()) {
```
hemberger, line 553: This logic can be simplified with something like
```
return <statement>
```
instead of
```
if (statement)
   return true
else
   return false
```

lib/Default/AbstractSmrPlayer.class.php
```diff
@@ -2214,6 +2227,9 @@ public function killPlayer($sectorID) {
 		$this->setNewbieTurns(100);
 	}
 
+	/*
```
hemberger, line 2230: Happy to add this unrelated comment, but might as well make it javadoc like the rest (double asterisk on the first line).

lib/Default/AbstractSmrPlayer.class.php
```diff
@@ -2339,9 +2369,10 @@ public function &killPlayerByPlayer(AbstractSmrPlayer $killer) {
 
 			$killer->increaseHOF($return['BountyGained']['Amount'], array('Killing', 'Money', 'Bounty Gained'), HOF_PUBLIC);
 
-			if ($this->getShip()->getAttackRatingWithMaxCDs() <= MAX_ATTACK_RATING_NEWBIE && $this->hasNewbieStatus() && !$killer->hasNewbieStatus()) { //Newbie kill
+			if ($this->isNoncombatNewbie() && !$killer->hasNewbieStatus()) { //Newbie kill
 				$killer->increaseHOF(1, array('Killing', 'Newbie Kills'), HOF_PUBLIC);
-			} else {
+				$killer->increaseCurrentBountyAmount('HQ', $killer->getExperience() * NEWBIE_KILL_BOUNTY_FACTOR);
```
hemberger, line 2374: Need to use `self::` to refer to class constants.

lib/Default/AbstractSmrPlayer.class.php
```diff
@@ -542,6 +544,17 @@ public function isNPC() {
 	public function hasNewbieStatus() {
 		return $this->newbieStatus;
 	}
+	
+	/**
+	 * Does the player have newbie status and
+	 * a low attack rating?
+	 */
+	public function isNoncombatNewbie() {
```
hemberger, line 552: Since every call to this function also asks if the killer is not in newbie status, I think it would help consistency if this function (renamed, perhaps) took a `SmrPlayer $killer` and included that logic, with a comment explaining that newbies are exempt from the penalties to killing other newbies.

lib/Default/AbstractSmrPlayer.class.php
```diff
@@ -2239,12 +2261,23 @@ public function &killPlayerByPlayer(AbstractSmrPlayer $killer) {
...
+			$killer->increaseExperience($return['KillerExp']);
 
-		$return['KillerCredits'] = $this->getCredits();
-		$killer->increaseCredits($return['KillerCredits']);
+			// Killer gains credits carried
+			$return['KillerCredits'] = $this->getCredits();
+			$killer->increaseCredits($return['KillerCredits']);
+			
+			// War setting gives them military pay
+			if ($relation <= RELATIONS_WAR) {
```
hemberger, line 2277: `$relation` is not defined until below (see Scrutinizer issue). You'll need to move that definition up here.

lib/Default/AbstractSmrPlayer.class.php
```diff
@@ -2222,11 +2238,17 @@ public function &killPlayerByPlayer(AbstractSmrPlayer $killer) {
...
-		$msg .= ' was destroyed by ' . $killer->getBBLink();
-		if ($killer->hasCustomShipName()) {
-			$named_ship = strip_tags($killer->getCustomShipName(), '<font><span><img>');
-			$msg .= ' flying <span class="yellow">' . $named_ship . '</span>';
+		//adjust message for newbie kill
+		if ($this->isNoncombatNewbie() && !$killer->hasNewbieStatus()) {
+			$msg .= ' was newbie killed by ' . $killer->getBBLink();
+		} else {
+			$msg .= ' was destroyed by ' . $killer->getBBLink();
+			if ($killer->hasCustomShipName()) {
```
hemberger, line 2246: Ship name logic doesn't need to be (and probably shouldn't be) inside the `else` clause.

hemberger (changes requested):
Thanks for the updates! A few issues need to be resolved, and then I'll take a final pass over it.

lib/Default/AbstractSmrPlayer.class.php
stupidnewbie: unrelated in terms of the PR, correct?  ideally all the functions would have documentation?

hemberger: If the behavior of a method is fully and clearly described by the method name, then I don't think adding a doc is helpful (this especially applies to trivial getters/setters). I also think it's better to have no docs than docs that don't add any new information, e.g.
```php
/**
 * Get the start sector
 */
function getStartSector {
  ...
}
```
But yes, in general I am a proponent of thorough documentation. :)

templates/Default/engine/Default/trader_examine.php
```diff
@@ -19,6 +19,9 @@
 	$canAttack = true;
 	$fightingPlayers = $ThisSector->getFightingTraders($ThisPlayer, $TargetPlayer, true);
 	if (count($fightingPlayers['Defenders']) > 0) {
+		if ($TargetPlayer->isNoncombatNewbie() && !$ThisPlayer->hasNewbieStatus()) {
```
hemberger, line 22: This still needs to be updated to the new function.

witold-cyrek:
The combat does not let itself to reading such warnings.  100% as the examine screen is loading the person is spammign the F key :)

PR #896, opened 2020-09-25
new shield weapons

stupidnewbie:
add two new anti shield weapons:

P3, Resonant Shield Disruptor, 110 shield damage at 65% accuracy
P4, Harmonic Shield Disruptor, 125 shield damage at 62% accuracy

There are currently only 2 neutral shield weapons that do not require an alignment selection to equip.  Neither of them over power 2.  By contrast, there are 11 armor weapons of the same nature ranging from P1 to P4.  This is a puzzling and somewhat frustrating part of the game world.

This will reduce the value of Thev/WQ/Nij/Creo relations as they currently carry all the reasonable shield weapons.  But their weapons are still better than these.

### DIFF
--- a/db/patches/V1_6_65_X__neutral_shield_weapons.sql
+++ b/db/patches/V1_6_65_X__neutral_shield_weapons.sql
@@ -1,0 +1,4 @@
+-- Add two new neutral anti-shield weapons.  One P3 and one P4.
+INSERT INTO `weapon_type` (`weapon_type_id`, `weapon_name`, `race_id`, `cost`, `shield_damage`, `armour_damage`, `accuracy`, `power_level`, `buyer_restriction`) VALUES
+(56, 'Resonant Shield Disruptor', 1, 99000, 110, 0, 65, 3, 0),
+(57, 'Harmonic Shield Disruptor', 1, 199000, 125, 0, 62, 4, 0);


hemberger:
The disparity between shield/armour weapons is quite jarring. I've been perusing the weapon list quite a lot lately, and I agree that some change like this is necessary, and your suggestions fit an unfilled niche. However, I want to be sure that in an effort to provide balanced weapon offers, we don't further degrade the usefulness of racial peace. Let's think on this a bit, maybe there's some more tweaking we can do.

stupidnewbie:
A more grand solution would be to revamp weapons and make racial ones better than the neutral counterparts.  For example, they could deal 5 or 10 percent more DPS.  Or put the racial weapons in lower accuracy brackets so they have higher burst and cap out at higher levels.  But that's pretty major surgery.

stupidnewbie:
One other small tweak solution occurred to me: tweak the human harmonic disruptor.  possibly make it level 3.  somewhere around 175 damage and 42% accuracy like a HPT.  it would be a sought after weapon with those stats.  versus now where it's widely avoided by experienced players.

There is a perceived advantage to lower accuracy/higher damage weapons.  I'm not sure if it's just folklore and superstition.  But I do I believe there's an accuracy pocket between about 25% and 70% where a weapon is broadly useful.  Lower than that and the weapon tends not to connect when you need it because a capital ship battle is only about 6 shots.  Higher than that and you start losing level bonuses because you hit 100% accuracy.

PR #897, opened 2020-09-27
add some convenience features to port attack warning screen

stupidnewbie:
add sector to port examine screen header

display attackers for port examine like a fleet battle would show

another one from accepted feature requests

### DIFF
--- a/engine/Default/port_attack_warning.php
+++ b/engine/Default/port_attack_warning.php
@@ -8,7 +8,7 @@ if ($sector->getPort()->isDestroyed()) {
 	forward(create_container('skeleton.php', 'port_attack.php'));
 }
 
-$template->assign('PageTopic', 'Port Raid');
+$template->assign('PageTopic', 'Port Raid In Sector #' . $player->getSectorID());
 
 $template->assign('PortAttackHREF', SmrSession::getNewHREF(create_container('port_attack_processing.php')));
 $template->assign('Port', $sector->getPort());
--- a/templates/Default/engine/Default/port_attack_warning.php
+++ b/templates/Default/engine/Default/port_attack_warning.php
@@ -51,3 +51,45 @@ if ($Port->isUnderAttack()) { ?>
 	<?php echo format_time($Port->getReinforceTime() - TIME) . '.';
 }
 ?></span>
+
+<?php
+if (!$canAttack) {
+	$fightingPlayers = $ThisSector->getPotentialFightingTraders($ThisPlayer);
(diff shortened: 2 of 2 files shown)


hemberger:
Thanks for the PR!

I'd like to get this merged, but since I've always disliked the "Examine" screen format (in particular because it's so verbose that I can't imagine it has ever been useful to anyone who didn't end up in a pod -- it's even difficult to count the number of attackers/defenders if there's more than two ships) , let me suggest an alternative:

Let's revamp the "Examine" page to our liking, and then reuse it as a template for a consistent display in trader, port, and planet examine pages.

What would you like it to look like? I imagine something similar to the "Ships" table in the current sector page. Would it be useful if you could hover over the ship name and get the old-style "Examine" info in a tooltip?

stupidnewbie:
I am still hashing this out but I have a start:

Each side needs a header where the total fleet health is listed.  Number of ships, cloak willing, would also be a good addition.  Or an average ship health remaining.  Also average or total player levels.

On the individual side: The named rank should be reported as a level.  Shield/CD/Armor values should move to a charted display.  Perhaps similar to ports credits/upgrade/raid look.  Maybe blue shield bar, white CD bar, brown armor bar.  With max/full values above the charts.

One other thing if we are expanding the scope here: we should tack a header on to attack screens showing condensed scan info overhead.  Either fleet header info or individual data depending on the combat type.

hemberger:
Oh yes, I like it!

For the header, I'm thinking we have something like

| Number of Ships  | Total Rating  | Average Level  |
|---|---|---|
| 5  | 150 / 90  | 10  | 

And then like you said maybe some normalized bars showing remaining shields/cds/armor? That might be a bit too much free information though, I don't know. At the very least, I think it could allow you to make much more informed firing decisions.

It's possible that we could even include such a header in the combat results page itself.

stupidnewbie:
On the free information bit - we currently see things like shields 500-600 armor 300-400 drones 100-200 right now.  And the numbers fall as the ship is damaged.  Showing that in something like a bar graph would still be showing coarse information.  I'm thinking it would be faster to read/identify as an analog display is all.

Embedding that in to the combat pages as a dashboard would make a ton of sense to me.  It puts the scan info my ship collects in a convenient place for when I need it the most.

Tacking on a lowest ship defense display would also be awesome for PR/PB.  That way we don't have to spend so much time staring at the current sector making sure someone isn't forgetting to repair when we're trying to fire.

PR #911, opened 2020-11-04
initial changes to support unit testing via phpunit

### LINKED ISSUE #910: Increase test coverage for AbstractSmrAccount
The goal is to increase automated test coverage of critical systems relating to account management, and state. Hopefully this helps eliminate a lot of extra local testing.

XaeroDegreaz:
There is a simple test for retrieving accounts via account id using a mock mysql object

closes #910


composer.json
```diff
@@ -12,5 +12,9 @@
         "php": "^7.4",
         "phpmailer/phpmailer": "^6.1",
         "vanilla/nbbc": "^2.0"
+    },
+    "require-dev": {
+        "phpunir/phpunit": "9.1.4",
```
hemberger, line 17: Is `phpunir` a typo? Looks like a weird package that's a wrapper around `phpunit`? https://github.com/MaurizioMoreo/test

XaeroDegreaz: Wow, very strange -- I hadn't even noticed that. I let autocomplete work itself out, and I thought I saw `phpunit` :D I'm gonna make sure to change this to the official distro.

lib/Default/MySqlDatabase.class.php
```diff
@@ -9,16 +9,29 @@ abstract class MySqlDatabase {
...
 	public function __construct($dbName) {
 		if (!self::$dbConn) {
 			// Set the mysqli driver to raise exceptions on errors
 			if (!mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT)) {
 				$this->error('Failed to enable mysqli error reporting');
 			}
 
-			self::$dbConn = new mysqli(self::$host, self::$user, self::$password,
-			                           $dbName, self::$port, self::$socket);
+            $host = self::$host;
```
hemberger, line 20: It looks like the indentation level for these 4 lines is too small.

lib/Default/MySqlDatabase.class.php
```diff
@@ -9,16 +9,29 @@ abstract class MySqlDatabase {
...
+            $user = self::$host;
+            $password = self::$password;
+            $port = self::$port;
+
+			// The configuration can be overridden via PHPUnit configuration
+			if (defined("OVERRIDE_MYSQL_CONFIG")) {
+				$host = constant("OVERRIDE_MYSQL_HOST");
```
hemberger, line 27: For these 4 variables, I don't think we need to use the `constant` function. We can just use, e.g.
```php
$host = OVERRIDE_MYSQL_HOST;
```
Or is there a good reason to do it like this? If so, we should do it globally. (The only reason I'm aware of to use this is if the constant name is variable.)

hemberger:
I'm coming around to your idea of parsing `.env` instead of having a separate mysql secrets file. Would something like https://github.com/vlucas/phpdotenv allow us to avoid the override logic in `SmrMySqlDatabase`?

.travis.yml
```diff
@@ -6,3 +8,14 @@ git:
...
   - tests/phplint.sh
   - tests/strict_types.sh
+  - composer install
+  - mv .env.sample .env
+  - mv config/config.specific.sample.php config/config.specific.php
+  - mv config/SmrMySqlSecrets.sample.inc config/SmrMySqlSecrets.inc
+  - mv config/discord/config.specific.sample.php config/discord/config.specific.php
+  - mv config/irc/config.specific.sample.php config/irc/config.specific.php
+  - mv config/npc/config.specific.sample.php config/npc/config.specific.php
+  - docker-compose run mysql-integration-test
```
hemberger, line 18: Maybe these need to be run in detached mode, i.e. with `--detach`? Otherwise I think they will block and prevent the remaining commands from being executed.

lib/Default/MySqlDatabase.class.php
XaeroDegreaz: The constant names are variables driven by the `phpunit.xml` file, so they have to be retrieved this way. You can do it with raw constants, but if you look in the code, there will be warnings that they are unknown constants.

.travis.yml
XaeroDegreaz: Yeah I fixed them. I was using `docker-compose run` instead of `up -d`.

lib/Default/MySqlDatabase.class.php
XaeroDegreaz: Thanks, I think that was remnants from the spaces formatting

XaeroDegreaz:
> I'm coming around to your idea of parsing `.env` instead of having a separate mysql secrets file. Would something like https://github.com/vlucas/phpdotenv allow us to avoid the override logic in `SmrMySqlDatabase`?

Yeah I think you could make that work.  Like we discussed earlier, I would just make sure that it doesn't actually read them into the environment for security.

hemberger:
I see some Scrutinizer warnings along the lines of:

> test/SmrTest/lib/DefaultGame/AbstractSmrAccountIntegrationTest.php#L167
> ``$socialLogin`` of type ``Mockery\Mock`` is incompatible with the type ``SocialLogin`` expected by parameter ``$social`` of ``AbstractSmrAccount::getAccountBySocialLogin()``.

Obviously we don't want to compromise our method typehints for the sake of the test infrastructure, but is there a recommended way to address this sort of issue, or is a linter like Scrutinizer always going to have a problem with test code like this?

Some quick googling shows discussions about issues that look similar to me, but it seems like it requires a nuanced understanding of PHPUnit, which I currently don't have!

XaeroDegreaz:
Let me look into this -- The object passed is not strictly of the type being mocked, but rather a proxy. I don't think that it compromises the typehint integrity in such a way that would affect production code, but it is a little annoying to see that the types returned from Mockery are not statically typed. However, to be fair, PHP is not a statically typed language, so this sort of makes sense.

XaeroDegreaz:
Ok, I've updated the tests to use the `PHPUnit` mocks instead. Their typehinting annotations should be good (at least the errors are gone from my IDE).

PR #915, opened 2020-11-12
[WIP] Use a single config file for MySQL configuration properties

### LINKED ISSUE #914: Use a single configuration file to manage MySQL config properties
Instead of using the `SmrMySqlSecrets.inc`, we van get away with using the `.env` file to manage all MySQL configuration. This will allow us to have a single config file used for Docker Compose, PHPUnit, and the SMR game.

XaeroDegreaz:
Originally branched from #910 So it shows those commits here, however the diff should show only the changes between this branch and master.

closes #914


lib/Default/MySqlDatabase.class.php
```diff
@@ -1,37 +1,28 @@
...
-	// add configuration static members via traits
-	use SmrMySqlSecrets;
-
 	protected static $dbConn;
 	protected static $selectedDbName;
 	protected $dbResult = null;
 	protected $dbRecord = null;
 
-	public function __construct($dbName) {
+	public function __construct($dbName=null, MysqlProperties $mysqlProperties=null) {
```
XaeroDegreaz, line 9: This is an iterative step towards setting up IOC Dependency injection with [PHP-DI](https://php-di.org/doc/understanding-di.html) in order to make classes more testable.

test/SmrTest/BaseIntegrationSpec.php
```diff
@@ -9,29 +9,57 @@
...
+use MysqlProperties;
 use PHPUnit\Framework\TestCase;
+use RuntimeException;
 use Throwable;
 
 
 class BaseIntegrationSpec extends TestCase {
 	protected static mysqli $conn;
 	private static $defaultPopulatedTables = array();
+	private const MYSQL_CONNECTION_ATTEMPTS = 5;
```
XaeroDegreaz, line 21: These could also be configured in the phpunit.xml configuration class.

lib/Default/MySqlDatabase.class.php
```diff
@@ -43,7 +34,7 @@ public function __construct($dbName) {
 		}
 
 		// Do we need to switch databases (e.g. for compatability db access)?
-		if (self::$selectedDbName != $dbName) {
+		if ($dbName != null &&self::$selectedDbName != $dbName) {
```
hemberger, line 37: This doesn't work in the case where `self::$dbConn` was previously opened for a "history" database, because if `$dbName === null`, then it won't switch back to the `smr_live` database. I'm not exactly sure how to fix this without knowing the name of the `smr_live` database a priori, but it also depends on how we want to handle "history" databases in general.

hemberger (commented):
Thanks for the PR! This looks like a great start. There are some issues I've outlined below, but don't feel like you need to address them all yourself. If you don't have an opinion or preference for how to resolve the issue, I'm happy to dig into it myself. I'm very grateful for the work you've contributed and don't want to scare you away! :D

I've added a commit that is needed to mount the `.env` file into the game container. However, it does occur to me that we're expanding the scope of some of our secret configuration variables by including the `.env` file. Presently, that's just the `MYSQL_ROOT_PASSWORD` (and some non-sensitive configuration variables like `PHP_DEBUG`), but it's something to be aware of.

Before addressing any of my specific comments below, I should mention that this currently doesn't work correctly for the `Smr{12,Classic}HistoryMySqlDatabase` classes. These were conditionally used based on the presence of `$dbName_Smr12History` and/or `$dbName_SmrClassicHistory` variables in `SmrMySqlSecrets.inc` (which has been deleted here). Since they're interacting with archived databases, we could probably get away with hard-coding the database name in the class, and then using a boolean `.env` variable to specify if the database exists.

docker-compose.yml
```diff
@@ -102,7 +103,7 @@ services:
     mysql-integration-test:
         <<: *mysql-common
         ports:
-            - 3307:3306
+            - ${MYSQL_PORT}:3306
```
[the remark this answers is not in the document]
XaeroDegreaz: I have 3307 set up in `phpunit.xml` because I have another MySQL server running on my machine, and don't want to use it for integration testing (different version, and rather use the Docker container). By default, it should stay 3306, and only use 3307 when the integration tests kick in.

lib/Default/MysqlProperties.class.php
```diff
@@ -0,0 +1,71 @@
+<?php declare(strict_types=1);
+
+
+use Dotenv\Dotenv;
+
+
+class MysqlProperties {
```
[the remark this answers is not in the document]
XaeroDegreaz: Fair enough!

lib/Default/MysqlProperties.class.php
```diff
@@ -0,0 +1,71 @@
...
+	private string $databaseName;
+
+	public function __construct() {
+		$configArray = !isset($_ENV["MYSQL_CONFIG_FROM_ENVIRONMENT"])
+			? Dotenv::createArrayBacked(ROOT)->load()
+			: $_ENV;
+		$this->parseConfigArray($configArray);
+	}
+
+	private function parseConfigArray(array $array) {
```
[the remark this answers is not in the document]
XaeroDegreaz: That works for me, I think I may have initially had another idea at some point (like testing this), but I suppose I could test it all from the constructor -- it would just be a little more involved than testing a pure, discrete method with a single responsibility.

lib/Default/MysqlProperties.class.php
```diff
@@ -0,0 +1,71 @@
...
+			MysqlProperties::CONFIG_MYSQL_PORT => $port,
+			MysqlProperties::CONFIG_MYSQL_DATABASE => $this->databaseName,
+		] = $array;
+		$this->port = (int)$port;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getHost(): string {
```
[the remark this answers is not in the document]
XaeroDegreaz: I'm a big proponent of making structs immutable whenever possible. Unfortunately PHP doesn't have the `final` keyword for fields, so making them public would make them mutable. It's also easier to mock these out by using methods, and it's generally considered best practice to make things immutable when they can be.

lib/Default/MysqlProperties.class.php
```diff
@@ -0,0 +1,71 @@
...
+	private const CONFIG_MYSQL_DATABASE = "MYSQL_DATABASE";
+	private string $host;
+	private string $user;
+	private string $password;
+	private int $port;
+	private string $databaseName;
+
+	public function __construct() {
+		$configArray = !isset($_ENV["MYSQL_CONFIG_FROM_ENVIRONMENT"])
+			? Dotenv::createArrayBacked(ROOT)->load()
```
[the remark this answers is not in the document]
XaeroDegreaz: Hm, I will look into a way to make this work. The goal of this was to make the config parsing agnostic to the source of the data (either from the `Dotenv` parsing, or from the the `$_ENV` array provided elsewhere).

phpunit.xml
```diff
@@ -10,21 +10,24 @@
 	failOnWarning="true"
 	verbose="true">
 	<php>
-		<const
-			name="OVERRIDE_MYSQL_CONFIG"
+		<env
+			name="MYSQL_CONFIG_FROM_ENVIRONMENT"
```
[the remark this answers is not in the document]
XaeroDegreaz: Yeah we could have a separate `.env` file just for tests. Like `test.env`, or something. I would still need an environment variable to dictate which file to choose to parse, though. SO there would still need to be some switching happening. In a future version, I plan to use dependency injection of properties like this so a lot of this can actually go away. It's out of scope for this ticket, so I have these changes stashed away in another branch.

lib/Default/MySqlDatabase.class.php
XaeroDegreaz: I was thinking about this, too, and I looked at the other implementations of this database class, and they don't seem to be referenced anywhere in the code, so it sorta felt like to me that it did not matter for them (I was going to delete them actually). I think we can come up with something better in the future with some IOC techniques I plan to apply.

Unless I'm missing something?

lib/Default/MysqlProperties.class.php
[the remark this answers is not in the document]
XaeroDegreaz: I pushes commit 649c9fafbbb6fe41e824131ba36b185b0cf448d3 which should address this. Not sure if Github will show this pointing to an outdated commit or not.

phpunit.xml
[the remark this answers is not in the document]
XaeroDegreaz: Resolved in 649c9fafbbb6fe41e824131ba36b185b0cf448d3

lib/Default/MySqlDatabase.class.php
```diff
@@ -1,37 +1,28 @@
...
-	// add configuration static members via traits
-	use SmrMySqlSecrets;
-
 	protected static $dbConn;
 	protected static $selectedDbName;
 	protected $dbResult = null;
 	protected $dbRecord = null;
 
-	public function __construct($dbName) {
+	public function __construct($dbName=null, MySqlProperties $mysqlProperties=null) {
```
hemberger, line 9:
```suggestion
	public function __construct($dbName = null, MySqlProperties $mysqlProperties = null) {
```
Spacing for Scrutinizer.

lib/Default/MySqlDatabase.class.php
```diff
@@ -1,37 +1,28 @@
...
 		if (!self::$dbConn) {
+			if(!$mysqlProperties){
```
hemberger, line 11:
```suggestion
			if (!$mysqlProperties) {
```
Spacing for Scrutinizer.

lib/Default/MysqlProperties.class.php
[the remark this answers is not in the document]
hemberger: Good point. For a moment I was confusing PHP with c++, where you can have const class members but still modify them in the constructor. Since that doesn't actually exist in PHP, private attributes with getters is :+1: with me!

lib/Default/MySqlDatabase.class.php
hemberger: We never refer to these classes explicitly. Here's an example of how it's used:
https://github.com/smrealms/smr/blob/7a4bd93dbb42ef081f098cdccfe2fc56a7026643/engine/Default/game_play.php#L150-L152
I think hard-coding the database name will help, but that alone won't fix the "switching back" issue.

lib/Default/MysqlProperties.class.php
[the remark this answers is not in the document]
XaeroDegreaz: Fixed with new validation stuff

lib/Default/MysqlProperties.class.php
```diff
@@ -0,0 +1,82 @@
+<?php declare(strict_types=1);
```
hemberger, line 1: I think this file needs to be deleted now that you have `MySqlProperties`.

docker-compose.yml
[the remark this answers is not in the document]
XaeroDegreaz: This should be fixed now as the test environment stuff and the docker stuff are not being used in the same process anymore

lib/Default/MysqlProperties.class.php
XaeroDegreaz: Hmm, weird. I did rename it. Unfortunately, I think because Windows is not case-sensitive, it isn't tracking correctly in git. I will try to fix it.

XaeroDegreaz: I had to set `git config core.ignorecase false`, then change the name a little, then change it back and it triggered a delete for `MysqlProperties.class.php`. I think it's fixed -- good catch

lib/Default/MySqlProperties.class.php
```diff
@@ -0,0 +1,82 @@
...
+			MySqlProperties::CONFIG_MYSQL_HOST => $this->host,
+			MySqlProperties::CONFIG_MYSQL_USER => $this->user,
+			MySqlProperties::CONFIG_MYSQL_PASSWORD => $this->password,
+			MySqlProperties::CONFIG_MYSQL_PORT => $port,
+			MySqlProperties::CONFIG_MYSQL_DATABASE => $this->databaseName,
+		] = $array;
+		$this->port = (int)$port;
+	}
+
+	public static function determineConfigEnvironmentFile(): string {
```
hemberger, line 32: Are these two methods public solely for the purpose of accessing them in unit testing? If so, I still feel that the interface should reflect its intended use, but could be convinced otherwise. It looks like there is a way to unit test private methods without making them public -- what do you think about https://jtreminio.com/blog/unit-testing-tutorial-part-iii-testing-protected-private-methods-coverage-reports-and-crap/#targeting-private-protected-methods-directly?

hemberger, line 32: Are these two methods public solely for the purpose of accessing them in unit testing? If so, I still feel that the interface should reflect its intended use, but could be convinced otherwise. It looks like there is a way to unit test private methods without making them public -- what do you think about https://jtreminio.com/blog/unit-testing-tutorial-part-iii-testing-protected-private-methods-coverage-reports-and-crap/#targeting-private-protected-methods-directly?

lib/Default/MySqlProperties.class.php
```diff
@@ -0,0 +1,82 @@
...
+	public static function validateConfig(Dotenv $config) {
+		$config->required(self::CONFIG_MYSQL_HOST);
+		$config->required(self::CONFIG_MYSQL_USER);
+		$config->required(self::CONFIG_MYSQL_PASSWORD);
+		$config->required(self::CONFIG_MYSQL_PORT)->isInteger();
+		$config->required(self::CONFIG_MYSQL_DATABASE);
+	}
+
+	/**
+	 * @return string
```
hemberger, line 49: I think we can omit these docstrings, since they provide no information above and beyond the specified return type.

hemberger (changes requested):
I think we should consider removing the ability to specify the mysql port entirely. This may sound crazy at first, but hear me out :)

* Unless I'm missing something, I think the definition of `MYSQL_PORT` in the mysql docker environment is unused. It exposes port 3306 in its container, and this is hard-coded in the mysql Dockerfile.

* The docker-compose `port` label, when specified, exposes the port on the host. Previously, we've restricted access to the mysql container by putting it in the backend network.

* Rather than expose the mysql-integration-test container on the host, it can be uniquely accessed by setting the .env variable `MYSQL_HOST` to something different than you use in the non-test .env file.

If this all makes sense, then I think that I would prefer to remove the ability to specify the port entirely in MySqlProperties, and just use `null` as we were before.

lib/Default/MySqlProperties.class.php
XaeroDegreaz: tl;dr - Testing private and protected methods is an anti-pattern.

I feel like using reflection to circumvent access modifiers is very hacky, and that private / protected method business logic should be tested from the top-down (through the public interface accessing those methods).

However, in this case, there is no way to test that business logic from the top down. Lets say we put all of the contents in that constructor, how can you test that each branch is being hit for explicitly determining the file name for the environment? Also, the config validation requires an instance of the `Dotenv` backend, so we would have to basically pass that in as a dependency into the constructor in order to validate it.

IMHO, these methods are fine to be public; they don't mutate any state. Unfortunately, PHP does not have something like "package-private" like JAVA. If that were the case, I could just remove them from the public API, and as long as my test was in the same package I could test them that way.

lib/Default/MySqlProperties.class.php
XaeroDegreaz: Yeah, IDE generated these. I can remove no problem.

lib/Default/MySqlProperties.class.php
hemberger: For this codebase, I think I would prefer visibility to indicate intended use. Otherwise, it sounds like the goal would be to make every private method public -- _for testing_ -- and that doesn't feel right to me. If we really want to test it outside the public API, then there is still a way to do it, and I'd rather the hacks be in testing infrastructure than to compromise the API.

I'm definitely open to discussing it further!

hemberger: I'm happy to leave these as public, as you say they don't mutate any state. We're obviously coming at this from different angles, but I'm willing to see where this leads. :)

XaeroDegreaz: I can inject the `Dotenv` dependency in the constructor and use a spy in order to verify interactions of the object. The dependency injection container stuff is in the next phase of my refactoring, so can I just do it there?

hemberger: Sure, go for it! It sounds like your plans to use DI here will be the best of both worlds.

Since I may not have been clear about why I commented on the method visibility in the first place, let me add some context. (I'm not advocating or lobbying for any change here, just hoping to provide some additional solid ground for any future discussions on the topic. This may all be obvious to you, but I don't think it hurts to be explicit!)

Let's say I'm a new developer to the project, and I decide to work on the `MySqlDatabase` class. I see that I need to use a `MySqlProperties` object. I head over to https://smrealms.github.io/api-docs and look at the public interface, and see some straightforward getter methods, but also some methods like `validateConfig`. Since it's public, it would be reasonable to ask "Do I need to call that? If I don't, does the config not get validated?". Instead of the API being prescriptive, I have to look at the source code to check. I'll then see that it's already called in the constructor, so I can infer that I don't need to call it myself. But that's more work, and could become burdensome if the entire codebase is written that way.

I see (at least) 3 possible ways to resolve such an issue to make it easier for a developer to use one part of the code without needing to know the internals the classes it uses:

1. Make the method private (though this can impede unit testing).
2. Document the method so that it's clear if/when someone would need to call it (not prescriptive, but better than nothing!).
3. Redesign the class to avoid the issue in the first place (e.g. your DI efforts).

hemberger:
Following up on my comment about `MYSQL_PORT`, I'm not exactly sure how we want to resolve this. I think it depends on how you prefer to specify things for the test environment. In some sense, we're currently overloading the responsibility of the `MYSQL_{PORT,HOST}` variables:
* `MYSQL_PORT` refers to both the port in the container (which is hard-coded to 3306 by the mysql Dockerfile) and the port that it will be bound to on localhost, which is only relevant for running tests with the local PHP environment.
* `MYSQL_HOST` refers to both the name of the mysql container in the docker network and the name of the host that is being used to expose the mysql port locally.

Here's one possible solution:

Omit the use of `MYSQL_PORT` in docker-compose.yml, and retain `MYSQL_HOST` as the name of the host as it will be referred to within the docker network for, e.g., flyway and pma (in the `test.env` file, perhaps we set it to something like `smr-mysql-test`). Then add extra .env variables for setting up `MySqlProperties` with a local PHP environment, perhaps `LOCAL_MYSQL_{HOST,PORT}` or similar?

If this seems like a reasonable point, do you want to take a crack at modifying the setup? Happy to discuss further if it's unclear  what I'm trying to get at!

Dockerfile
```diff
@@ -46,6 +46,10 @@ RUN MODE=$([ "$PHP_DEBUG" = "0" ] && echo "production" || echo "development") &&
 	echo "Using $MODE php.ini" && \
 	tar -xOvf /usr/src/php.tar.xz php-$PHP_VERSION/php.ini-$MODE > /usr/local/etc/php/php.ini
 
+RUN if [ "$PHP_DEBUG" = "1" ] ; then pecl install xdebug \
+	&& echo "zend_extension=/usr/local/lib/php/extensions/no-debug-non-zts-20190902/xdebug.so" >> /usr/local/etc/php/php.ini ; \
```
hemberger, line 50:
```suggestion
	&& docker-php-ext-enable xdebug ; \
```
I think this is the recommended extension installation procedure. Documentation is at https://hub.docker.com/_/php under "PECL extensions".

.env.sample
```diff
@@ -4,10 +4,12 @@
...
 #   might need to remove "-" characters)
-#
-# NOTE: These variables must be consistent with config/SmrMySqlSecrets.inc
 
 MYSQL_ROOT_PASSWORD=smr
+MYSQL_USER=smr
 MYSQL_PASSWORD=smr
 MYSQL_HOST=smr-mysql
+MYSQL_DATABASE=smr_live
+MYSQL_PORT=3306
```
hemberger, line 13: It looks like `MYSQL_PORT` has been removed from this PR? If so, we should probably remove it here as well.

phpunit.xml
```diff
@@ -10,21 +10,9 @@
...
-			name="OVERRIDE_MYSQL_USER"
-			value="smr" />
-		<const
-			name="OVERRIDE_MYSQL_PASSWORD"
-			value="smr" />
-		<const
-			name="OVERRIDE_MYSQL_PORT"
-			value="3307" />
+		<env
+			name="MYSQL_CONFIG_ENVIRONMENT"
```
hemberger, line 14: Is the `MYSQL_CONFIG_ENVIRONMENT` variable being used anymore? It seems like no, since we're mounting `test.env` in the container as `.env`?

test/resources/mysql-config/validation/good.env
```diff
@@ -0,0 +1,8 @@
+MYSQL_ROOT_PASSWORD=smr
+MYSQL_USER=smr
+MYSQL_PASSWORD=smr
+MYSQL_HOST=127.0.0.1
+MYSQL_DATABASE=smr_live
+MYSQL_PORT=3306
```
hemberger, line 6: I think we can remove `MYSQL_PORT` here as well?

.travis.yml
```diff
@@ -6,16 +14,16 @@ php:
 git:
   depth: false
 script:
+  - docker-compose --version
   - test/phplint.sh
   - test/strict_types.sh
   - composer install
```
hemberger, line 20: Since we're now running PHPUnit inside the containerized PHP environment, do we still need to run `composer install`? I think you would know best, but if we don't need it then I guess we can save a little bit of time by omitting it.

hemberger (commented):
This PR is looking great! I've got a few minor cleanup suggestions/questions, but I think this is just about ready to merge. Thanks for all your hard work!

.travis.yml
XaeroDegreaz: Yeah I guess since the build of the container will do this, no need for it here.

Dockerfile
XaeroDegreaz: Cool; updated

lib/Default/MySqlProperties.class.php
```diff
@@ -0,0 +1,49 @@
...
+	private string $user;
+	private string $password;
+	private string $databaseName;
+
+	public function __construct() {
+		$config = Dotenv::createArrayBacked(ROOT);
+		$array = $config->load();
+		self::validateConfig($config);
+		[
+			MySqlProperties::CONFIG_MYSQL_HOST => $this->host,
```
hemberger, line 20: Okay, I think this is my last suggestion: Should we use `self::CONFIG_MYSQL_*` (instead of explicitly writing out the class name here)?

hemberger (approved):

PR #923, opened 2020-11-17
Upload test coverage report to codecov.io from Travis

hemberger:


hemberger:
@XaeroDegreaz How does this look to you? See details at https://codecov.io/gh/smrealms/smr/branch/codecov (it's not as useful as it can be since the master branch doesn't have any coverage reports yet).

I have the coverage options in the command line, because I couldn't figure out how to specify them in the `phpunit.xml` file. It looks like the syntax changed in a recent version, and the documentation and common knowledge hasn't caught up yet. Perhaps not ideal, but shouldn't really be an issue.

.travis.yml
```diff
@@ -24,3 +24,6 @@ script:
   - mv config/npc/config.specific.sample.php config/npc/config.specific.php
   - composer run start:integration-services
   - composer run test
+
+after_success:
+    - bash <(curl -s https://codecov.io/bash)
```
XaeroDegreaz, line 29: What is this doing?

hemberger: It uploads the output `test/coverage.xml` file to the smrealms/smr project on codecov.io. I'll add a comment saying so.

XaeroDegreaz: You're not specifying that file anywhere though o_O

XaeroDegreaz: Wait, it's downloading a script or something and then executing it in bash? How does it know where to grab the xml file?

hemberger: It's magic. I think it scans for valid coverage files. I was thinking about explicitly adding in the `-f tests/coverage.xml` option so it's a little more clear/explicit about what it's doing (and probably will do that now), but it's not necessary.

XaeroDegreaz: Oh ok

hemberger: Yeah, it's downloading a bash script.

Issue #920, opened 2020-11-16
Switch from php templating to Twig


hemberger:
We currently use PHP templating to render the display pages. We may want to consider incrementally migrating to Twig (https://twig.symfony.com/doc/) instead.

Benefits:
* One of the most standard PHP templating libraries.
* Enforces a serializable, read-only data model (whereas PHP templating still has access to our PHP libraries and can modify the state of existing objects and databases).
* The syntax is less verbose `{{ bla }}` vs. `<?php echo $bla; ?>` and has many convenient functions for manipulating the data (e.g. escaping etc.).

Downsides:
* It's an entirely new language that must be learned to maintain our codebase.
* It creates a dependency on an external library (though it is so popular that it is unlikely to be abandoned).
* The additional overhead likely would incur a small runtime penalty.

PR #924, opened 2020-11-17
Add dependency injection to MySqlDatabase, and MySqlProperties

### LINKED ISSUE #922: Start using PHP-DI to manage class dependencies starting with MySqlDatabase class
[PHP-DI](https://php-di.org/doc/understanding-di.html) Is used by [millions](https://packagist.org/packages/php-di/php-di) of PHP projects in order to manage "Inversion of Control (IOC)" dependency injection.

In this first iteration, the goal is to make it such that the `MySqlDatabase` class is ever only constructed once during a request. Currently it is constructed many times, and relies on static fields to hold maintain state. The DI container will manage the creation of the object, and inject any dependencies it might have. In future iterations, this class will also be injected into other service classes, like the `AbstractSmrAccount` class.

The goal is to start taking a cleaner approach to object lifecycles, and make all aspects of testing easier by ensuring that we can easily test each unit's business logic without having to worry about dependencies that are created during / _after_ the construction of an object.

XaeroDegreaz:
closes #922


admin/Default/changelog.php
```diff
@@ -6,7 +6,7 @@
 $template->assign('ChangeMessage', $var['change_message'] ?? '');
 $template->assign('AffectedDb', $var['affected_db'] ?? '');
 
-$db2 = new SmrMySqlDatabase();
+$db2 = MySqlDatabase::getInstance(true);
```
XaeroDegreaz, line 9: I wonder if adding a `MySqlDatabase::getNewInstance()` may be nice to have so that there is a little more clarity as to what's going on.

tools/irc/server.php
```diff
@@ -169,8 +169,8 @@ function server_msg_401($fp, $rdata)
 
 		echo_r('[SERVER_401] ' . $server . ' said: "No such nick/channel" for ' . $nick);
 
-		$db = new SmrMySqlDatabase();
-		$db2 = new SmrMySqlDatabase();
+		$db = MySqlDatabase::getInstance();
+		$db2 = MySqlDatabase::getInstance(true);
```
XaeroDegreaz, line 173: I'd fix this, but I want you to see scrutinizer's error before I do it, and then you wonder why I did it :D

test/SmrTest/lib/DefaultGame/MySqlDatabaseTest.php
```diff
@@ -0,0 +1,47 @@
...
+		$this->container = DiContainer::getContainer();
+		$this->mysql = $this->createMock(mysqli::class);
+		// Replace the factory definition for mysqli object with our mock, so when
+		// requesting a mysqli instance, it will always return a mock for this test
+		$this->container->set(mysqli::class, $this->mysql);
+	}
+
+	public function test__construct_happy_path() {
+		$this->mysql
+			->expects(self::once())
```
XaeroDegreaz, line 31: I'm working on a way to fix Scrutinizer on this, but I'm not sure how just yet. Unfortunately PHP7 doesn't seem to support union types. I tried fixing it with `@var` in the PHPDoc, but that didn't satisfy The Scrute.

I may end up switching up the testing infrastructure in favour of [PHPSpock's PHPUnit adapter](https://github.com/aleczhang/PhpSpock). Then I can write these tests like I write [Spock tests](http://spockframework.org/spock/docs/1.3/all_in_one.html) in the JVM world. Spock tests are really slick, and easy to read without near as much setup.

hemberger: PHP8 will have union types, and it is released in just over a week (Nov 26, 2020). I will be upgrading the containers to PHP8 as soon as I can. Will that solve this problem?

Looking at the PhpSpock library, it's barely been touched since it was created 9 years ago (and your PR is the only PR ever!). I'm not sure I would be comfortable adding a dead dependency.

tools/irc/server.php
hemberger: Yep, feel free to remove that line.

admin/Default/changelog.php
hemberger: I have yet to review the new interface, but my gut reaction is that a separate, explicit method would be nice. Alternatively, this type of interface would be a candidate for named arguments, which are coming in PHP8.

lib/Smr/Container/DiContainer.php
```diff
@@ -0,0 +1,112 @@
...
+				return new mysqli(
+					$mysqlProperties->getHost(),
+					$mysqlProperties->getUser(),
+					$mysqlProperties->getPassword(),
+					$mysqlProperties->getDatabaseName());
+			},
+			Dotenv::class => function (): Dotenv {
+				return Dotenv::createArrayBacked(ROOT);
+			},
+			// Explicitly name all classes that are autowired, so we can take advantage of
```
hemberger, line 55: I'm a bit confused about these class enumerations. One of the benefits of autoloader is that you only load the files you need. If we use DI throughout the entire library, and we have to explicitly specify them here, then that means we'll be loading all the files as soon as we load the `DiContainer`. (And I suspect this wouldn't work for classes that use the game-type dependent loading, e.g. `lib/Default` vs. `lib/Draft`, which is kind of a hack, so maybe this will open the door for a better design for that.)

Now, I do think that we eventually want to take advantage of preloading in at least the production image (already available in PHP 7.4). That might make my concern about overzealous autoloading moot.

lib/Smr/Container/DiContainer.php
```diff
@@ -0,0 +1,112 @@
...
+	 * @throws \DI\DependencyException
+	 * @throws \DI\NotFoundException
+	 */
+	public static function make(string $className) {
+		return self::getContainer()->make($className);
+	}
+
+	/**
+	 * Return the raw Container instance
+	 * @return Container
```
hemberger, line 107: I'd say let's get rid of all the docstring elements that offer no documentation above and beyond the interface itself.

lib/Smr/Container/DiContainer.php
```diff
@@ -0,0 +1,112 @@
...
+			 * mysqli is a 3rd party library, and we do not have control over its constructor.
+			 * In order for PHP-DI to construct an instance of a class, each constructor argument must
+			 * be able to be constructed.
+			 * PHP-DI cannot construct mysqli by itself, because all of its arguments are primitive types.
+			 * Therefore, we need to declare a provider factory for the container to use when constructing new instances.
+			 *
+			 * The factories themselves are able to use dependency injection as well, so we can provide the MysqlProperties
+			 * typehint to make sure the container constructs and instance and provides it to the factory.
+			 */
+			mysqli::class => function (MySqlProperties $mysqlProperties): mysqli {
```
hemberger, line 41: Can this anonymous function be moved to a static function of `MySqlDatabase`? I think that if a developer wants to understand how we interact with the database, they should only really need to look at the `MySqlDatabase` class. The `DiContainer` class is probably the last place I'd expect to find it. :)

Does that seem reasonable, or is there a good argument for keeping it here?

hemberger (commented):
This looks okay in general so far, though the benefits are not manifestly obvious to me yet. I see how the containers are being used in the tests, and that is nice, but I feel like I've done the same sort of mocking in python, for example, without needing IoC/DI (since these techniques are sometimes too heavyweight for simple programs).

Practically speaking, the MySqlDatabase class is now certainly more "pure", but it is also less clear to me how it interacts with the rest of the codebase. The static mysqli attribute in comparison is quite straightforward.

The design looks solid though. Happy to continue with this effort and see what we look like on the other side!

lib/Default/MySqlDatabase.class.php
```diff
@@ -1,33 +1,43 @@
...
+
+	/**
+	 * MySqlDatabase constructor.
+	 * Not intended to be constructed by hand. If you need an instance of MySqlDatabase,
+	 * use MySqlDatabase::getInstance();
+	 * @param mysqli $dbConn The mysqli instance
+	 * @param MySqlProperties $mysqlProperties The properties object that was used to construct the mysqli instance
+	 */
+	public function __construct(mysqli $dbConn, MySqlProperties $mysqlProperties) {
+		$charset = $dbConn->character_set_name();
```
hemberger, line 34: The check on the character set is a sanity check on the mysqli object, not on the MySqlDatabase object, and it should only be performed once per connection. Therefore, it should probably be moved to the `mysqli::class` "maker" (what's the correct terminology for that?).

test/SmrTest/lib/DefaultGame/MySqlDatabaseTest.php
XaeroDegreaz: Union types should solve the problem, I think.

Another issue is that the base PHPUnit mocking/stubbing does not lent itself well to the refactoring of method names. In the expectation chain you declare a method to be called. If this method gets renamed, or moved, the tests to not reflect the change because you are using a string literal.

With PhpSpock, the method / fields that are stubbed out / verified are always going to be correct, as the object that gets returned is a correctly typehinted object (not some mock implementation). I don't know if you've ever had the pleasure of working with Spock for unit tests in the past, but it's the cleanest, most efficient testing framework I've ever used.

If something works, and works well it doesn't really need to be updated. As long as it fulfils its purpose, I have no problem with using old dependencies (especially a test dependency). The only issue that I've experienced with PhpSpock is that the PHPUnit namespaces are old, but I resolved them locally and created several tests using their adapter, and it worked gloriously.

lib/Smr/Container/DiContainer.php
```diff
@@ -0,0 +1,112 @@
...
+			},
+			// Explicitly name all classes that are autowired, so we can take advantage of
+			// the compiled container feature for a performance boost
+			MySqlProperties::class => autowire(),
+			MySqlDatabase::class => autowire()
+		];
+	}
+
+	private function buildContainer(): Container {
+		$builder = new ContainerBuilder();
```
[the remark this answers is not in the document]
XaeroDegreaz: The `Container` manages object dependencies for you. That means creating them when needed, and supplying them to dependents when requested.

Please see the documentation for [compiling the container](https://php-di.org/doc/performances.html). In essence, it only happens one time per deployment. I've enabled that caching feature as sort of a pre-optimization. It exists inside the `/tmp` directory on the Docker container itself, so you don't have to worry about removing the compiled output when redeploying.

lib/Smr/Container/DiContainer.php
XaeroDegreaz: You don't have to specify them, I just added them explicitly so that the definitions can be compiled. Also, everything that I've defined in here are already loaded in every page request -- the `MySqlDatabase` is constantly being used everywhere. So we're not loading anything that doesn't need to be.

lib/Smr/Container/DiContainer.php
XaeroDegreaz: All it's doing is returning a `mysqli` object -- all the other interactions are already happening in the `MySqlDatabase` class. This is a factory method for producing the `mysqli` dependency required by the database class to operate. I'd be willing to add some more documentation around the constructor for the `MySqlDatabase` class, but I think this is the appropriate place for this dependency factory to be. It's not immediately obvious in this iteration, but there is a purpose for this being here :D

lib/Default/MySqlDatabase.class.php
XaeroDegreaz: The term is "factory".

I can move it to the factory and adjust the tests.

admin/Default/changelog.php
XaeroDegreaz: Yeah but the named arguments might be even more wordy :D I'm also sort of leaking implementation details here, so perhaps concealing those is a better option

lib/Smr/Container/DiContainer.php
XaeroDegreaz: Hm, I mean I think it's good to document what it's actually returning. Like, what `Container`? An explanation of the container it's returning seems useful to me. Maybe reword it? Any thoughts?

lib/Default/MySqlDatabase.class.php
XaeroDegreaz: Actually, now that I think about it, the charset check is outside the scope of the factory method. That check only takes place once the object is fully constructed. By that point, the factory has completed its responsibility. This means that this implementation detail, or business logic is the responsibility of the dependent who's using it -- the dependent determines whether or not they want to use the state of the dependency.

Factories should be dumb, and just "give you a thing". They shouldn't contain nearly any business logic. An exception to this rule is if you need some business logic to determine HOW to create a dependency, or if a dependency is even eligible for construction. For instance, this is valid factory business logic:

```php
if (!mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT)) {
	throw new RuntimeException('Failed to enable mysqli error reporting');
}
```

lib/Smr/Container/DiContainer.php
XaeroDegreaz: Also about the game depend loading -- let's cross that bridge when we get there. There are several more iterations to go through before this thing starts to jump out at you and provide an "aha" moment.

lib/Smr/Container/DiContainer.php
hemberger: After doing some testing with phpdoc, I'm kind of annoyed that it doesn't infer the type from the typehint. I dislike duplicating programmatic information in the docstring, because it increases the cost of maintaining it and the likelihood of desynchronization. Anyway, that's not something we can solve here, so I'm happy with this docstring as-is. :)

XaeroDegreaz: Are you talking about how auto completion isn't working when using this method or something?

hemberger: No, just the phpDocumentor web pages (https://smrealms.github.io/api-docs/).

lib/Smr/Container/DiContainer.php
hemberger: Okay, sounds good.

test/SmrTest/lib/DefaultGame/MySqlDatabaseTest.php
XaeroDegreaz: Are you vehemently opposed to me using PhpSpock? I will use my fork so it's updated

hemberger: The benefits of Spock sound nice. I feel similarly about `unittest` vs. `pytest` when I'm working in python. I am still worried about the fact that the project is abandoned, because everything abandoned breaks at some point. I guess the big questions are: a) who is going to migrate back to PHPUnit when it breaks and b) how much effort will it be to do that? If the answers are "@hemberger" and "a lot", then I would definitely want to stick with PHPUnit. What do you think?

XaeroDegreaz: It's still using PHPUnit

hemberger: I saw that PhpSpock had PHPUnit as a dependency. Can you show me an example of what the tests would look like with & without PhpSpock?

lib/Smr/Container/DiContainer.php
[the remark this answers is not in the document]
hemberger: Thanks for the information. The code comments say much the same thing, but sometimes it takes a few hits over the head for me to get it. :)

From the PHP-DI docs, I see:
> When you deploy new versions of your code to production you must delete the generated file (or the directory that contains it) to ensure that the container is re-compiled.

So I think we want to turn this off in the `smr-dev` container, since the source code is volume-mounted there to allow us to change it without rebuilding the image. (I would prefer not to overload the `PHP_DEBUG` flag for this purpose, since I often turn it off in my `smr-dev` container. Instead, maybe we can hard-wire some variable in the docker-compose configuration in the `smr` and `smr-dev` service environments.)

test/SmrTest/lib/DefaultGame/MySqlDatabaseTest.php
XaeroDegreaz: It doesn't have a dependency on PHPUnit -- you have to provide that yourself if you want to use the adapter:

Example (base PHP):
```php
public function test_validate_config_happy_path() {
	//# Given a Dotenv object
	$dotEnv = $this->createMock(Dotenv::class);
	// And the dotenv config will return the following array when loaded
	$dotEnv
		->expects(self::once())
		->method("load")
		->willReturn([
			"MYSQL_HOST" => "host",
			"MYSQL_USER" => "user",
			"MYSQL_PASSWORD" => "pass",
			"MYSQL_DATABASE" => "database"
		]);
	// And we expect that the dotenv "required" method will be called with the following arguments
	$dotEnv
		->expects(self::exactly(4))
		->method("required")
		->with(
			$this->logicalOr(
				"MYSQL_USER",
				"MYSQL_HOST",
				"MYSQL_PASSWORD",
				"MYSQL_DATABASE"));
	// When constructing the properties class
	$mysqlProperties = new MySqlProperties($dotEnv);
	// Then the properties have expected values
	self::assertEquals("host", $mysqlProperties->getHost());
	self::assertEquals("user", $mysqlProperties->getUser());
	self::assertEquals("pass", $mysqlProperties->getPassword());
	self::assertEquals("database", $mysqlProperties->getDatabaseName());
}
```

Example Spock:
```php
public function test_validate_config_happy_path_using_spock() {
	/**
	 * @var $dotEnv \Dotenv\Dotenv *Mock*
	 */
	setup:
	1 * $dotEnv->load() >> array(
		"MYSQL_HOST" => "host",
		"MYSQL_USER" => "user",
		"MYSQL_PASSWORD" => "pass",
		"MYSQL_DATABASE" => "database");
	1 * $dotEnv->required("MYSQL_USER");
	1 * $dotEnv->required("MYSQL_HOST");
	1 * $dotEnv->required("MYSQL_PASSWORD");
	1 * $dotEnv->required("MYSQL_DATABASE");
	when:
	$mysqlProperties = new MySqlProperties($dotEnv);
	then:
	$mysqlProperties->getHost() == "host";
	$mysqlProperties->getUser() == "user";
	$mysqlProperties->getPassword() == "pass";
	$mysqlProperties->getDatabaseName() == "database";
}
```

lib/Smr/Container/DiContainer.php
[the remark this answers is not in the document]
XaeroDegreaz: Yeah good point! I will work on something

hemberger (changes requested):
This is looking great! I have some minor suggestions & requests, but I think there's also one bug that needs to be sorted out. Please see my comments below, and let me know if you have any questions. Thanks!

lib/Default/MySqlDatabase.class.php
```diff
@@ -37,42 +48,42 @@ public function __construct() {
...
 	}
 
 	// This should not be needed except perhaps by persistent connections
 	public function close() {
-		if (self::$dbConn) {
-			self::$dbConn->close();
-			self::$dbConn = false;
+		if ($this->dbConn) {
+			$this->dbConn->close();
+			unset($this->dbConn);
```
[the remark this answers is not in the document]
XaeroDegreaz: Try `getNewInstance`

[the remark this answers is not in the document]
hemberger: Are you sure that works? If I understand correctly, a new instance of `MySqlDatabase` will still be using the `mysqli` instance from the DI container, and thus the same connection, which will have been closed. Also, in this case, it won't necessarily be known where the next call to `getInstance` would be (it may be a place that, under most conditions, should use the existing `MySqlDatabase` instance, and thus can't be replaced with `getNewInstance`).

I realize the original design of `MySqlDatabase` is probably contributing to some of these problems (since it conflates the idea of connection vs. result, as one of its many flaws). It's not yet clear to me if additional refactoring is needed to solve this, or if there's a simple solution like the one you proposed.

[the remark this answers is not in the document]
XaeroDegreaz: You're right I was in a hurry to reply on mobile and didn't think it through. I will look in the codebase and see where close is getting called and try to find an appropriate way to solve the issue.

I feel like views should not be accessing the database layer like that and having control over low level service routines like that. Also I can't think of a reason to close and reopen a connection to a database in the same page request. That being said I will look into a way to resolve this but yes I agree more refactoring is required. Eventually my goal is to remove accessing that class directly for any view objects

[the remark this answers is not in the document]
hemberger: Currently, this method is only used by the CLI programs, which need to close the connection between callbacks to avoid timeout errors (see 319c684eef). Like you, I can't think of a reason it would be needed for page requests.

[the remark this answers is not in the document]
XaeroDegreaz: I also noticed that the `close()` method is used in `mysql_cleanup.php`. That `mysql_cleanup` function is referenced in a few script (they look like mostly Discord stuff).

I guess my question is, is there really a need to create new connections after the connection is closed? Is something in these CLI scripts opening and closing multiple times?

How necessary is it to have the ability to open another connection after it's been closed?

I guess we could have something equivalent to:

```php
public function query($query) {
	if (!$this->dbConn) {
		DiContainer::getContainer()
			->set(
				mysqli::class,
				DiContainer::make(mysqli::class));
		$this->dbConn = DiContainer::get(mysqli::class);
	}
	$this->dbResult = $this->dbConn->query($query);
}
```

Could be extracted to some sort of utility method somewhere, but I feel like maybe this is an edge case that doesn't really need to be resolved. I could be misunderstanding everything that's going on during the CLI calls, but if there really is ever a need for this edge case to be resolved we could circle back and do something about it then? Otherwise it looks like we'd be implementing something that wouldn't be used.

lib/Default/MySqlDatabase.class.php
```diff
@@ -1,33 +1,44 @@
...
-			// Default server charset should be set correctly. Using the default
-			// avoids the additional query involved in `set_charset`.
-			$charset = self::$dbConn->character_set_name();
-			if ($charset != 'utf8') {
-				$this->error('Unexpected charset: ' . $charset);
-			}
+use Smr\Container\DiContainer;
+use Smr\MySqlProperties;
+
+class MySqlDatabase {
```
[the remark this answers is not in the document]
XaeroDegreaz: Yeah I think I will probably only move this later on. Limit scope for now

lib/Default/MySqlDatabase.class.php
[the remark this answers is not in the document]
hemberger: It is currently essential to the proper functioning of the IRC and Discord bots that they be able to close the mysql connection between callbacks, because otherwise the connection times out and the bots no longer respond to user commands until they are restarted. If there is a way to solve this problem while keeping the mysql connection open, that would alleviate the need for a `close` method. I'm not a database expert, but I spent quite a long time debugging this back in 2017 and I couldn't find any other solution.

lib/Smr/Container/DiContainer.php
```diff
@@ -0,0 +1,115 @@
...
+		}
+		return $builder->build();
+	}
+
+	/**
+	 * Create a new DI\Container instance.
+	 * This needs to be done once during a bootstrapping script, like htdocs/config.inc
+	 */
+	public static function initializeContainer(): void {
+		new DiContainer();
```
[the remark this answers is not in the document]
XaeroDegreaz: I guess I'm just used to setting up fields in the constructor, but since this is static, I guess it doesn't matter. Changed

lib/Default/MySqlDatabase.class.php
```diff
@@ -1,33 +1,44 @@
...
+	private mysqli $dbConn;
+	private MySqlProperties $mysqlProperties;
+	private string $selectedDbName;
+	/**
+	 * @var bool | mysqli_result
+	 */
+	private $dbResult = null;
+	private ?array $dbRecord = null;
+
+	public static function getInstance(bool $requireNewInstance = false): MySqlDatabase {
```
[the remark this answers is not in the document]
XaeroDegreaz: Are you sure `this::class` is legal? I get warnings in my IDE, and tests also do not pass.

```
require(/smr/htdocs/../lib/Default/this.class.php): failed to open stream: No such file or directory
```

`self::class` seems to work, though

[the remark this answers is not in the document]
hemberger: Sorry about that. Yes, I meant `self::class`. Good catch!

lib/Smr/Container/DiContainer.php
XaeroDegreaz: It's a dependency of the class, so why should it not be injected as a dependency? It makes it near impossible to do simple tests otherwise. If I want to test failure scenarios, or business logic on that dependency it cannot be done easily, or concisely.

I feel like you can clearly see how this class is using the `mysqli` dependency, and I don't know how having it's construction right in this class helps understanding anything extra; there's only one way to construct that thing :D

New developers who don't understand about dependency injection techniques can just ask questions, or we can have more documentation in the readme about the paradigm shift happening in the codebase.

I also feel like "I don't understand it" is not a valid reason to skimp out on modern best practices for software architecture, but if you feel so strongly about this then I'm willing to concede the whole thing.

However, I can't guarantee you the test coverage the codebase deserves without using this tried-and-true pattern.

lib/Default/MySqlDatabase.class.php
XaeroDegreaz: How would you test that branch of logic if it happened in the factory method?

lib/Smr/Container/DiContainer.php
hemberger: I'm not advocating for a change in the dependency injection scheme, just moving the guts of the factory method to a static function of `MySqlDatabase` so that it's the one-stop-shop for seeing how our class wraps the `mysqli` library.

Let me lay out a sketch of what I'd like to see. For example, in `DiContainer.php`, we'd have:
```php
mysqli::class => function (MySqlProperties $mysqlProperties): mysqli {
    return MySqlDatabase::createMysqliObject($mysqlProperties);
}
```
And in `MySqlDatabase.class.php`, we'd have:
```php
/**
 * This method will be used by the DI factory to create the mysqli object, etc.
 */
public static function createMySqliObject(MySqlProperties $mysqlProperties): mysqli {
    // Set the mysqli driver to raise exceptions on errors
    if (!mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT)) {
        throw new RuntimeException('Failed to enable mysqli error reporting');
    }
    return new mysqli(
        $mysqlProperties->getHost(),
        $mysqlProperties->getUser(),
        $mysqlProperties->getPassword(),
        $mysqlProperties->getDatabaseName());
    // And if I were writing this, I'd put the utf-8 sanity check here so that it is performed once per connection. YMMV!
}
```

Functionally, I don't think this is any different than what you're doing. Organizationally, you may still dislike it, but I believe it should still allow you to perform all the testing you want to do.

Does this clarification change the equation at all for you?

lib/Default/MySqlDatabase.class.php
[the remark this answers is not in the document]
XaeroDegreaz: I see, ok so they are long-lived processes these CLI guys. Hmm... Yeah well the `ping` command used to reconnect to the database, but looks like that is sorta a deprecated function now that PHP uses `mysqlnd`.

I think the only way to make this work is to implement some reconnect logic like above until one of the later refactors in which nothing uses `MySqlDatabase` directly.

lib/Default/MySqlDatabase.class.php
hemberger: I would test it by mocking the `mysqli` constructor.

lib/Smr/Container/DiContainer.php
XaeroDegreaz: Yeah, it's a compromise I can live with. But now you have one of those pesky `public static` methods hanging around

lib/Default/MySqlDatabase.class.php
XaeroDegreaz: How can you mock something you don't have control of though? It's being constructed inside the method -- you don't have access to mock that guy's constructor at that point.

XaeroDegreaz: Example:

Given this factory method:
```php
function factory(MySqlProperties $mysqlProperties): mysqli {
    if (!mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT)) {
        throw new RuntimeException('Failed to enable mysqli error reporting');
    }
    $mysqli = new mysqli(
        $mysqlProperties->getHost(),
        $mysqlProperties->getUser(),
        $mysqlProperties->getPassword(),
        $mysqlProperties->getDatabaseName());
    //# You need to mock this call
    $charset = $mysqli->character_set_name();
    //# In order to test this branch
    if ($charset != 'utf8') {
        throw new RuntimeException('Unexpected charset: ' . $charset);
    }
    return $mysqli;
}
```

From this test function:
```php
function test_non_utf_throws_exception() {
    //# Given mysql properties mock
    $mysqlProperties = MockWWithAllPropertiesSet();
    //# When producing a mysqli object through the factory method
    $mysqli = factory($mysqlProperties); //# How can you mock mysqli? This instantiates a real object.
}
```

lib/Default/MySqlDatabase.class.php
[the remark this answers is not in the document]
hemberger: Perhaps here isn't the place, but could you give me some insight into your end-game with `MySqlDatabase`? What still uses this class? What do other parts of the code use instead? How would this redesign allow us to close and reopen mysql connections in an intuitive way? It sounds like you have it planned out in your head, so maybe you can share some spoilers. :)

This snippet above looks like it probably works if the next call is `$db->query(...)`, but obviously `query` isn't the only method that uses the `dbConn`. Maybe this stopgap needs to go into `getInstance` (and `getNewInstance`?) instead?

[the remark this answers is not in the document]
XaeroDegreaz: About the location, I was just quickly spitballing something.

Ok so this might be long:

The end goal here is to separate concerns, encapsulate the code architecture into "domains of responsibility", and promote statelessness where possible.

It's a wide-sweeping thing that I'm _trying_ to do in functional (in the case of the `MySqlDatabase` class, temporary) pieces, and prevent breakage by preserving backwards compatibility. It started off with `MySqlDatabase`, since lots of things are touching it, but I plan on moving onto something like the "Account Domain" next.

One thing I'm trying to prevent is the `view` layer of the codebase from accessing the database directly. Instead, the views should use a `service` layer in order to do such things. Right now, IMHO, there is too much implementation detail about `MySqlDatabase` being leaked at too high of a level. Many views perform a query, check if there are results, iterate on those results, and other things. Those things could be handled in a service dedicated to that domain.

Want information about an account, or player? Use the account or player service. Want some voting information? Use the voting service.

You already have the `AbstractSmrAccount` class which is a combination of a `service`, `repository`, and a `model`/`entity`/`DAO`. I think we both agreed that this was pretty a "overloaded" class. It seemed the next logical step to work in since it doesn't seem like it gets updated all that often, so me separating that out into its constituent parts wouldn't interfere as much.

I'm also trying to promote "statelessness" in these classes. For instance, a big problem with `MySqlDatabase` class is that it has state for the `record` and the `result` of the queries. One of the longer-term goals is to remove that responsibility from this wrapper. We've talked, and you've mentioned, that there are conflation issues with this class. Not only is it a service, but it's also acting as a partial repository AND a stateful query result wrapper.

All of this leads back to a cleaner code structure, and one that is _easily_ testable whether it's an integration test for database repositories, or simple unit tests.

Here's a contrived example modern pattern I'm working to achieve.

```php
class AccountService {
    private AccountRepository $accountRepository;

    public function __construct(AccountRepository $accountRepository) {
        $this->accountRepository = $accountRepository;
    }

    public function findByEmail($email): Account {
        $account = $this->accountRepository->findByEmail($email);
        // ... Do some processing and return results
        return $account;
    }
}

class AccountRepository {
    private MySqlDatabase $mySqlDatabase;

    // In all actuality, I would want to remove many of the guts of MySqlDatabase at some point
    // So it can truly be a stateless class
    public function __construct(MySqlDatabase $mySqlDatabase) {
        $this->mySqlDatabase = $mySqlDatabase;
    }

    public function findByEmail($email): Account {
        // ... Write query, and return results, if any
    }
}

class Account {
    public $username;
    public $email;
}

// some_account_page.php
$account = $accountService->findByEmail("asd@asd.asd");
```

Let's note that the `findByEmail` sort of looks like a passthru, but the `AccountService` class would of course contain many more useful business methods than that. However I hope the comments inside the method show their responsibilities even though they share a common name.

I offered once, not sure if you remember or saw it, but we could create an architectural diagram of the current way things look, and an end goal. We can also get together and define these "domains of responsibility", and the way to approach migrating things over safely little-by-little.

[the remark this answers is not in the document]
XaeroDegreaz: Some definitions:
**View layer**: The layer that the user sees. Should be responsible for displaying data retrieved from some service authority.
**Service layer**: This is the layer that performs all of the heavy lifting and business logic, and returns output to controller layers, and view layers. This layer is probably the most important, and should contain as much test coverage as possible.
**Repository layer**: This layer is responsible for querying a data store, and returning results

lib/Default/MySqlDatabase.class.php
hemberger: Yep it's a good point. I guess something like [this](https://akrabat.com/replacing-a-built-in-php-function-when-testing-a-component/) (leveraging namespace resolution rules to override the global namespace)? In python I'd use a patch decorator, but it doesn't look like PHP(Unit) has that type of functionality.

Happy to go with the new design, and maybe in some iteration down the road there will be a way to get the best of both worlds.

XaeroDegreaz: Mockery does have some support for [global mocks](http://docs.mockery.io/en/latest/reference/creating_test_doubles.html#overloading), and that was actually the first mocking framework I tried to use for testing `AbstractSmrAccount` without spinning up a full integration test. Unfortunately it has some drawbacks when needing to do stuff like this multiple times in the same test suite. It also required stuff like running `PHPUnit` tests in separate processes so namespace/class collisions wouldn't happen.

With that you lose the ability to set breakpoints in classes :( It was just not a good solution, but a solution that does indeed work when you have these types of "hard dependencies" being constructed in the middle of your test subject methods, and have no alternative.

lib/Default/MySqlDatabase.class.php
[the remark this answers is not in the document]
hemberger: Thanks for the insightful comments. It's useful to have a bit of concrete code to see where you're trying to get to.

Translating some of these terms to our existing infrastructure:

**view layer** is our `templates/` (though there's a bit too much database/object access in there, which is part of what https://github.com/smrealms/smr/issues/920 is trying to remedy).

**service layer** is our `engine/` (though again too much direct database access -- ideally it would just be using Smr library functions/classes).

And our `lib/` is pretty much a mishmash of everything else. :P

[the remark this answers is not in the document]
XaeroDegreaz: Updated so that `getInstance()` and `getNewInstance()` will reconnect when needed

lib/Smr/Container/DiContainer.php
XaeroDegreaz: Moved

lib/Default/MySqlDatabase.class.php
```diff
@@ -14,11 +14,51 @@ class MySqlDatabase {
...
+	 * Verifies that the MySqlDatabase instance has a valid $dbConn instance,
+	 * and rewires the DI container to use freshly instantiated instances in the event
+	 * of disconnect from the database.
+	 * @param MySqlDatabase $mysqlDatabase
+	 * @return MySqlDatabase
+	 * @throws \DI\DependencyException
+	 * @throws \DI\NotFoundException
+	 */
+	private static function ensureConnected(MySqlDatabase $mysqlDatabase): MySqlDatabase {
+		if (!isset($mysqlDatabase->dbConn)) {
```
XaeroDegreaz, line 34: Now that I'm thinking about this, I can probably just get away with calling `DiContainer::initialize()` to reset the whole container. If not, then in the future after things are retrieving services through the container, they could end up with a defucnt transient `MySqlDatabase` instance if the connection is ever closed. That would be a lot of manual checks everywhere to make sure reconnect logic is triggered.

I think it's not a big deal since only those CLI apps are using the `close` method, but once the migration starts happening where they don't access db directly, but through domain services, this could be a problem. Maybe just reset the entire DI container for those CLI apps instead of having this reconnect logic.

hemberger: What you have set up now looks like it reproduces the existing behavior. But I agree that avoiding a defunct `MySqlDatabase` instance is probably more robust (especially if future abstraction layers would make misuse easier). However, a few questions I don't have answers for off the top of my head:

What would `DiContainer::initialize()` do to the existing connection? Would it try to close it (if so, might this throw if the connection has timed out)? Or would it leave it open (if so, might this result in a proliferation of open database connections)?

Also, it's important to keep in mind that not all CLI events even need a connection to the database, so I'm not exactly sure where a `DiContainer::initialize()` would go (if we're trying to be as efficient as possible). There are plenty of "reasonable" places to put it, though!

So I'm happy to leave this as-is for now, but if you want to look for a better solution, that's fine with me too.

lib/Smr/Container/DiContainer.php
```diff
@@ -0,0 +1,105 @@
...
+		return [
+			/*
+			 * mysqli is a 3rd party library, and we do not have control over its constructor.
+			 * In order for PHP-DI to construct an instance of a class, each constructor argument must
+			 * be able to be constructed.
+			 * PHP-DI cannot construct mysqli by itself, because all of its arguments are primitive types.
+			 * Therefore, we need to declare a provider factory for the container to use when constructing new instances.
+			 *
+			 * The factories themselves are able to use dependency injection as well, so we can provide the MysqlProperties
+			 * typehint to make sure the container constructs and instance and provides it to the factory.
```
hemberger, line 37:
```suggestion
			 * The factories themselves are able to use dependency injection as well, so we can provide the MySqlProperties
			 * typehint to make sure the container constructs an instance and provides it to the factory.
```
Just fixing some typos.

hemberger:
This is merged now. Thanks for your contribution (and your patience)!

PR #931, opened 2020-12-01
soften neutral relations grind

stupidnewbie:
Currently relations gain logic for neutral matches racial.  The game mechanic is 1 relation point gained per 30 holds up to 300 holds for a max of 10 relations gained up until relations hit 500.  After which the gain is capped to 1 relation per trade.  This makes perfect sense for racial space as it provides a home court advantage and makes political relations paramount to efficient inter racial trading.

The same reasoning does not follow with neutral relations.  There is no political relation mechanic here.  The result is a short grind to 500 and a miserable grind to 1000.

This commit singles out neutral relations in the increase/decrease logic and changes the way they scale.  New model is: one relation gained per 30 cargo traded up to 150 cargo.  That means a max of 5 relations gain (or loss when an error occurs) per trade.  And the cap at 500 is simply removed.

tl;dr: neutral relations grow at half the speed of racial relations but never cap at 1 per trade.

### DIFF
--- a/lib/Default/AbstractSmrPlayer.class.php
+++ b/lib/Default/AbstractSmrPlayer.class.php
@@ -1493,10 +1493,14 @@ abstract class AbstractSmrPlayer {
 	 * of the port given by $raceID.
 	 */
 	public function increaseRelationsByTrade($numGoods, $raceID) {
-		$relations = ICeil(min($numGoods, 300) / 30);
-		//Cap relations to a max of 1 after 500 have been reached
-		if ($this->getPersonalRelation($raceID) + $relations >= 500) {
-			$relations = max(1, min($relations, 500 - $this->getPersonalRelation($raceID)));
+		if ($raceID == RACE_NEUTRAL) {
+			$relations = ICeil(min($numGoods, 150) / 30);
+		} else {
+			$relations = ICeil(min($numGoods, 300) / 30);
+			//Cap relations to a max of 1 after 500 have been reached
+			if ($this->getPersonalRelation($raceID) + $relations >= 500) {
+				$relations = max(1, min($relations, 500 - $this->getPersonalRelation($raceID)));
+			}
 		}
 		$this->increaseRelations($relations, $raceID);
 	}
@@ -1506,7 +1510,11 @@ abstract class AbstractSmrPlayer {
 	 * bargaining and getting caught stealing.
 	 */
 	public function decreaseRelationsByTrade($numGoods, $raceID) {
-		$relations = ICeil(min($numGoods, 300) / 30);
+		if ($raceID == RACE_NEUTRAL) {
+			$relations = ICeil(min($numGoods, 150) / 30);
+		}else{
+			$relations = ICeil(min($numGoods, 300) / 30);
+		}
 		$this->decreaseRelations($relations, $raceID);
 	}
 


hemberger:
Thanks for putting something together for this!

I have a few concerns about the proposal:

1. If you trade a neutral route regularly, your relations will be absurdly high (since there is no soft cap), and in such a case there will never be a punishment for activities that degrade neutral relations.

2. At the start of the game, for all races besides your own, you can't get good bargains with other races until political action is taken (i.e. first you can't trade at all, and then you get soft capped at 800 total relations after peace goes through). With no neutral soft cap, neutral ports essentially become proxies for ports of your own race, except they take 2-4x as many trades to master. In some sense this is a nerf to Alskant.

3. There should be some downside to trading a neutral port, since it has the advantage that it can be traded by all races. Up until recently (714c7b5d80e68f42779cb720734317e34ba196cd), you couldn't even raise neutral relations at all, so you were stuck trading at 0 relations. After that change, you got soft capped at 500 relations, so you had to keep at it to get perfect relations, so the downside was that you couldn't speed it up with peace.

Is there anything more creative we can do with the neutral ports, perhaps? I've always wanted to covert "Neutral" to a droid race, but that's probably a bit extreme. ;)

stupidnewbie:
1: I had not thought of it that way.  Perhaps a more reasonable solution is to adjust the soft cap?  Either by raising it or perhaps capping gains at 2 or 3?

2/3: There's a downside via opportunity cost inherent to trading neutral.  At the start of a round your turns are nearly always better spent trading with your own race.  And later on neutral relations are very cost prohibitive to build.  I believe 50.5 turns hours in a freighter is the minimum time cost to build neutral 500 -> 1000.  That alskant coffin will do it a little faster but a PSF takes 125.5 hours so it's really 2-5 days minimum to go 500->1000 in a trade ship.

It does makes sense to me having neutral trade be a bit stunted compared to trading at home or using political relations effectively.  At the same time a barrier of 2-5 days worth of turns strikes me as cripplingly high.  Might a slower overall growth rate and a soft cap around 750 or 800 be a reasonable compromise?

On the creative front, we could bind neutral ports to an entirely different trade formula.  Calling them androids would be a reasonable lore justification for whatever insanity we could conjure up :)  Maybe have low relations improve prices less than high ones.  Perhaps run them on a 2000 or 3000 relations cap with a long term higher profitability.  Something along those lines would give them a niche at least.

witold-cyrek:
I think the end game should be that the port gets raced or everyone takes ages to grind it out :)  Also there is an easier way to get the relations, by buying 1 item and gaining 1 relation for 1 turn. By dumping 300 turns you can go from 500 to 800 relations, which is good enough to trade quite well.  Until that is changed this seems like unneeded change.

hemberger:
I think Astax has a good point, so I don't see a strongly compelling reason to make a change here. Also, from a code design/maintenance perspective, I like when we can avoid special cases (proliferation of if-statements makes testing harder).

I'm still open to a more substantive redesign, but if we want to pursue that, let's move the conversation to https://github.com/smrealms/smr/discussions or https://github.com/smrealms/smr/issues.

I'll close this PR for now, but please feel free to reopen it if you feel like this is premature. Either way, I'm very grateful for the thought that went into this PR -- thanks newbie!

PR #945, opened 2020-12-07
Translate between SMR and RouteGenerator transaction perspectives

Page-:
RouteGenerator code assumes transactions from port perspective whilst
SMR assumes transactions from player perspective, this moves the
conversion into the `GOOD_BUYS`/`GOOD_SELLS` constants so the rest of
the code can ignore the difference

### DIFF
--- a/lib/Default/Routes/OneWayRoute.class.php
+++ b/lib/Default/Routes/OneWayRoute.class.php
@@ -66,15 +66,16 @@ class OneWayRoute extends Route {
 
 	public function getMoneyMultiplierSum() : int {
 		// Relations factors assume maximum relations
-		$sellRelFactor = 3; // Note that this assumes maximum relations
-		$buyRelFactor = 1;
+		$buyRelFactor = 3; // Note that this assumes maximum relations
+		$sellRelFactor = 1;
 		// Supply factors assume maximum supply
-		$sellSupplyFactor = 2;
-		$buySupplyFactor = 1;
+		$buySupplyFactor = 2;
+		$sellSupplyFactor = 1;
 		$goodInfo = \Globals::getGood($this->goodId);
-		$buyPrice = IRound(0.03 * $goodInfo['BasePrice'] * pow($this->buyDi, 1.3) * $buyRelFactor * $buySupplyFactor);
-		$sellPrice = IRound(0.08 * $goodInfo['BasePrice'] * pow($this->sellDi, 1.3) * $sellRelFactor * $sellSupplyFactor);
-		return $sellPrice - $buyPrice;
+		$sellPrice = IRound(0.03 * $goodInfo['BasePrice'] * pow($this->sellDi, 1.3) * $sellRelFactor * $sellSupplyFactor);
+		$buyPrice = IRound(0.088 * $goodInfo['BasePrice'] * pow($this->buyDi, 1.3) * $buyRelFactor * $buySupplyFactor);
+		// NOTE: RouteGenerator code uses buy/sell from the perspective of the PORT not the player, hence why this seems the wrong way round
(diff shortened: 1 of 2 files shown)


hemberger (changes requested):
If I recall correctly, the main impetus for this change was to make the translation between the LazyTrader code and the SMR RouteGenerator easier. However, I think I'd prefer to keep SMR internally consistent as much as possible, to avoid the confusion that results from mixing the perspectives. Would that be okay with you?

If in the future we decide that it makes more sense to consistently use the _port_ perspective, then hopefully #979 will make that easier to do.

There are still a few vestiges of the backwards implementation that was fixed in 01f83ecdd70d73466c8a32da2c613a666fe4c214. In particular, the RouteGenerator passes the _current_ sector as the "sell sector" and the _target_ sector as the "buy sector". This is probably the reverse of what most people would expect, but in practice I don't think it matters since the current/target labels are arbitrary in the way that we're using the RouteGenerator class. If we decide we want to switch that, I'm happy to help. Otherwise, please feel free to close this PR if you don't think any further action is required.

Thanks!

Page-:
@hemberger having the buy/sell mixed up is going to be breaking the generation of money routes, which means npcs are going to be trading bad money routes and therefore never earn enough money to get into better ships - that was the reason for the change, to make npcs work correctly again

hemberger:
Can you provide the SMR file where you're seeing this issue? I've done a lot of testing, and I haven't yet encountered any discrepancies between TLT and the SMR RouteGenerator.

For example, here's a comparison of the output between the two. As I mentioned, the designation of the "current" and "target" sectors results in the reversed ordering of the route string (i.e. starts at sector B instead of sector A), but otherwise the profit calculations are identical.

TLT:
```
Exp: 1.04	Money: $173.12
Route: 1 (Neutral) buy Narcotics at 7x to sell at (Distance: 10) 37 (Alskant) at 10x
37 (Alskant) buy Luxury Items at 1x to sell at (Distance: 10) 1 (Neutral) at 7x

Exp: 1.23	Money: $171.38
Route: 1 (Neutral) buy Narcotics at 7x to sell at (Distance: 9) 26 (Neutral) at 9x
26 (Neutral) buy Luxury Items at 4x to sell at (Distance: 9) 1 (Neutral) at 7x

Exp: 1.56	Money: $170.13
Route: 1 (Neutral) buy Narcotics at 7x to sell at (Distance: 7) 60 (Ik'Thorne) at 7x
60 (Ik'Thorne) buy Luxury Items at 7x to sell at (Distance: 7) 1 (Neutral) at 7x

Exp: .96	Money: $139.37
Route: 1 (Neutral) buy Narcotics at 7x to sell at (Distance: 10) 37 (Alskant) at 10x
37 (Alskant) buy Weapons at 1x to sell at (Distance: 10) 1 (Neutral) at 5x

Exp: 1.14	Money: $135.00
Route: 1 (Neutral) buy Narcotics at 7x to sell at (Distance: 9) 26 (Neutral) at 9x
26 (Neutral) buy Weapons at 4x to sell at (Distance: 9) 1 (Neutral) at 5x
```

SMR RouteGenerator:
image

Page-:
Ah I could have sworn they would have worked out differently :|  Still it'd be nice if they could have identical output (and ideally keep the code as similar as possible to allow sharing bugfixes - the main benefits of TLT is it can run standalone against a .smr and is multi-threaded/heavily optimized for speed when generating multi-port routes)

hemberger:
Agreed, for sure! I think with a minor tweak to the RouteGenerator (e.g. swapping the current/target sectors etc.), the "forward" and "return" routes should show up in the same order as TLT. I'll see if I can work that out, but the impact should only be cosmetic.

PR #947, opened 2020-12-09
Added Plot.ly graphs for Racial Strengths

witold-cyrek:
Added new graphs using Plotly.  This is the best I can do, as it is unwieldy beast :)


htdocs/js/smr15.js
```diff
@@ -102,8 +102,57 @@ function showRaceInfo(select) {
 	var desc = document.getElementById('race_descr');
 	$(desc).children().addClass('hide');
 	$(".race_descr" + race_id, desc).removeClass('hide');
+	graph_by_plotly(race_id);
 }
 
+// Used by game_join.php 
+function graph_by_plotly(race_id) {
```
hemberger, line 109: I think this could use a more descriptive function name, e.g. `createRaceRadarChart` or similar (camel case for consistency).

htdocs/js/smr15.js
```diff
@@ -102,8 +102,57 @@ function showRaceInfo(select) {
...
+
+	var races = [["[1.0,1.0,1.0,1.0,1.0]","'Neutral'","'#FFD800'"],
```
hemberger, line 111: A few minor issues with the `races` data structure:
1. Formatting
    * Add spaces between items (i.e. between the values, the race name, and the color).
    * I don't know what the name of the style is, but for consistency let's format arrays like
```
array = [
   [[1.0,1.0,1.0,1.0,1.0], 'Neutral', '#FFD800'],
   ...
];
```
2. There is a typo with `Alskant`. I don't like duplicating the names and colors here, because duplication leads to human error (typos, desync, etc.). While it would be pretty easy to pass the race names in from PHP, the colors would be challenging, because they only exist in the CSS files, and I don't know of a simple way to generate the CSS files from PHP with our infrastructure.
3. Dummy 0th index should be removed. I know it's annoying that the races are 1-indexed instead of 0-indexed. Perhaps it would be better to use an associative array, and then the keys can be the actual `race_id` without a dummy first entry, e.g.
```
var races = {
  1: [[1.0,1.0,1.0,1.0,1.0], 'Neutral', '#FFD800'],
  2: ...
};
```

htdocs/js/smr15.js
```diff
@@ -102,8 +102,57 @@ function showRaceInfo(select) {
...
+	[[1.0,1.0,1.0,1.0,1.0],'Neutral','#FFD800'],
+	[[5.2, 10.0, 5.8, 8.1, 5.2],'Alaskant','#FF00FF'],
+	[[8.3, 5.0, 10.0, 3.3, 8.3],'Creonti','#FF8000'],
+	[[9.5, 4.9, 9.5, 7.7, 9.5],'Human','#0000FF'],
+	[[8.0, 6.1, 9.7, 10.0, 8.0],'Ik\'Thorne','#BFBFFF'],
+	[[8.1, 4.7, 8.8, 4.3, 8.1],'Salvene','#00AA00'],
+	[[10.0, 5.7, 8.7, 9.1, 10.0],'Thevian','#800000'],
+	[[7.2, 6.3, 7.9, 6.2, 7.2],'WQ Human','#804040'],
+	[[9.5, 4.9, 8.0, 5.1, 9.5],'Nijarin','#FF8080']];
+	data = [
```
hemberger, line 121: You have a couple Scrutinizer warnings here. They should be pretty easy to fix (e.g. using `var data` instead of just `data`, etc.).

htdocs/js/smr15.js
```diff
@@ -102,8 +102,57 @@ function showRaceInfo(select) {
...
+		{
+			type: 'scatterpolar',
+			r: races[race_id][0],
+			theta: ['Hunting','Trading','Combat', 'Utility', 'Hunting'],
```
hemberger, line 125: Why is "Hunting" in this array twice? Can this be fixed by just omitting one of them, or is this a hack to solve some other issue?

(Also, spaces between items.)

templates/Default/engine/Default/game_join.php
```diff
@@ -6,7 +6,7 @@
 	<h2>Round Description</h2>
 	<p><?php echo bbifyMessage($Game->getDescription()); ?></p><?php
 }?>
-
+<script src='https://cdn.plot.ly/plotly-latest.min.js'></script>
```
hemberger, line 9: Two minor issues:
1. We are trying to adopt a consistent style of using double quotes in HTML (instead of single quotes).
2. Let's fix to a specific version of plotly so that we can control when we update to a new library (otherwise the graph could break when a new version is released).

templates/Default/engine/Default/game_join.php
```diff
@@ -151,17 +151,12 @@
...
 					<tr>
-						<td colspan="3">Trading</td>
-					</tr>
-					<tr>
-						<td>Combat<br />Strength</td>
 						<td>
-							<img width="440" height="440" id="graphframe" src="images/race/graph/graph<?php echo $SelectedRaceID; ?>.gif" alt="Race overview" />
+							<div style="width: 550px; height: 550px;background: noone;" id="graphframe"></div>
+							<script>
+								graph_by_plotly('<?php echo $SelectedRaceID?>');
```
hemberger, line 157: JavaScript can't be included mid-HTML anymore. To see why, open this page in Chrome, open the inspector and turn off caching (to exaggerate the effect), and then reload the page over and over again. What you'll see is that the page rendering is interrupted while the JS is being run, and since this occurs in the middle of a table, it momentarily breaks the page formatting.

See d0ec278efebc8ad845eb15e25fc74ad4dbfe6741 for how I solved this problem with our list.js implementation. Unfortunately, that doesn't support completely generic JS execution, so I'm not sure what the best way to solve this problem is. (Do we create another if-clause specific to running this one function? That's going to get pretty out of control if we continue to add more JS, but we need to get it into `EndingJavascript.inc` somehow.)

hemberger (changes requested):
Thanks for taking a stab at this! I think the overall implementation is good, and there are just a few visual and coding tweaks that are needed to polish this off.

For reference, here's what the PR currently looks like:
image

Looking at the page as a whole, there's either too much whitespace around the graph, or everything else is too compressed. For the sake of maintaining the existing feel (for now), I think the best thing to do is remove the table around the graph and instead just use a centered div (the table was only used to set up the axis labels previously).

As you pointed out, the white background is bland, and looks scientific instead of sci-fi. We're going to need to spice that up a bit. If you think you've exhausted your patience with the Plotly docs, then I can browse around to see if I can find anything.

Lastly, it would be awesome if we could animate the graph, like the original gifs (growing from the center) or like the d3js example I linked (morphing between different states). This may be too much to ask for the first iteration, though, so I'm happy to move forward with this without animation.

Please see code comments below. Thanks again!

htdocs/js/smr15.js
witold-cyrek: This is required because last plot is from Utility back to Hunting.  It is how Plotly works. Omitting last plot leaves a missing line.

templates/Default/engine/Default/game_join.php
hemberger: It's not just this script tag, it's also the placement of `<script src='https://cdn.plot.ly/plotly-latest.min.js'></script>`. Both of these needs to go inside `EndingJavascript.inc`. Take out both and you'll see that the page renders properly.

hemberger:
I found how to specify the background color. It's an option of the `polar` element, e.g.
```
layout = {
  polar: {
    radialaxis: { ... },
    bgcolor: #XXX,
  }
}
```
I'm not sure what color it _should_ be. Probably something close to black, since the racial colors are so varied. Maybe like this?
image

witold-cyrek:
There is no way to reduce whitespace around the graph other than to hide the legend, which I did now.  The graph needs the area it needs. Also removing the table does very little, just risks CSS troubles on some resolution I have not tested, but I have removed it and tried my best to make the graph not break the page.

Also, there is no way to animate the graph as it states in the docs "Currently, only scatter traces may be smoothly transitioned from one state to the next. Other traces are compatible with frames and animations but will be updated instantaneously." So I take that to confirm what I have observed tryign to animate this thing. Only thing I can do is make the change between different graphs delayed by increasing frame duration, but the results appear instatly.  There could be a long winded work-around for this, animating each 'frame' manually using js callbacks, but that leads to all kinds of problems if someone switches the race dropdown while the animation is taking place.

htdocs/js/smr15.js
```diff
@@ -102,8 +102,78 @@ function showRaceInfo(select) {
...
 	$(desc).children().addClass('hide');
 	$(".race_descr" + race_id, desc).removeClass('hide');
+	createRaceRadarChart(race_id);
 }
 
+// Used by game_join.php 
+function createRaceRadarChart(race_id) {
+
+	var races = {
+		1: [[1.0,  1.0,  1.0,  1.0,  1.0],  'Neutral',    '#FFD800'],
```
hemberger, line 112: Since we're using an associative array now, there's no reason to have an entry for "Neutral" anymore. Let's just remove it.

htdocs/js/smr15.js
```diff
@@ -102,8 +102,78 @@ function showRaceInfo(select) {
...
+				color: '#fff',
+			},
+			bgcolor: '#111'
+		},
+		paper_bgcolor: '#06240E',
+		plot_bgcolor: '#000',
+		font: {
+			color: '#fff'
+		},
+		coloraxis : '#000'
```
hemberger, line 153: We can add the following element to `layout` to reduce the vertical whitespace:
```
margin: { l: 170, r: 170, t: 0, b: 0 }
```
(format this however you see fit)

templates/Default/engine/Default/game_join.php
```diff
@@ -146,25 +145,8 @@
...
-					</tr>
-					<tr>
-						<td colspan="3">Utility</td>
-					</tr>
-				</table>
-			</td>
-		</tr>
 	</table>
+	<div class="center">
+		<div style="width: 600px; height: 600px;background: none; margin: auto;" id="graphframe"></div>
```
hemberger, line 150: The `margin` change I recommended above allows us to remove all the `style` elements here.
```suggestion
		<div id="graphframe"></div>
```

hemberger (changes requested):
Yeah, the animation is definitely more trouble than it's worth at this point. Let's go back to the previous setup and remove the `animate` call.

A few minor visual changes below, and then I think we're good to go. Thanks!

hemberger (approved):

hemberger:
This is merged now. Thanks for your contribution!

PR #949, opened 2020-12-10
add scouts to backyard drone farm

stupidnewbie:
Currently scouts are only sold at a Combat Accessories during normal games.  This makes scouts unnecessarily rare as the CA locations are tightly controlled in order to keep mines away from fed beacons.  Furthermore, placing a CA in a place accessible to racial galaxies but not in a place to be mined in is a time consuming pain in the ass.

We could bypass that SNAFU by simply adding scout drones to the drone farm inventory.  Which this PR does!  (assuming no errors were committed!)


hemberger (approved):
Yep, let's give this a try in beta. I think it should provide a lot more flexibility in designing maps. Furthermore, I believe more accessibility to scout drones is likely to increase interaction.

PR #951, opened 2020-12-13
Always create new instances of MySqlDatabase

XaeroDegreaz:
In order to defeat global `$db` bug

### DIFF
--- a/admin/Default/changelog.php
+++ b/admin/Default/changelog.php
@@ -6,7 +6,7 @@ $template->assign('ChangeTitle', $var['change_title'] ?? '');
 $template->assign('ChangeMessage', $var['change_message'] ?? '');
 $template->assign('AffectedDb', $var['affected_db'] ?? '');
 
-$db2 = MySqlDatabase::getNewInstance();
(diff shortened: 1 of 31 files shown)


hemberger:
@XaeroDegreaz So it seems I can't request you as a reviewer for your own PR, but this is ready for you to review if you'd like to do so before it's merged. Thanks again for helping with this fix!

XaeroDegreaz:
Looks good dude. Wonder why the code coverage dropped. I guess there are a few things that are untestable now with the refactor, otherwise it looks like all the branches that were added are covered.

hemberger:
It was a little tricky to see where the coverage dropped (getting used to reading Codecov again!), but I think it was just the following two things:
1. The charset check that we moved from the `MySqlDatabase` ctor to the mysqli factory (which you warned would be untestable).
2. `DiContainer::get` is no longer called anywhere (except in the test setup). Probably needs to be added to the DiContainer unit test, perhaps? Not exactly sure how that should be tested though.

Thanks for taking a look at the changes!

hemberger (approved):

PR #958, opened 2020-12-30
Perform file inspection tests in phpunit

hemberger:
This allows the phplint and strict_type tests to be easily run in the containerized PHP environment.

### DIFF
--- a/tools/irc/irc.php
+++ b/tools/irc/irc.php
@@ -1,4 +1,3 @@
-#!/usr/bin/php -q
 <?php declare(strict_types=1);
 
 class TimeoutException extends Exception {}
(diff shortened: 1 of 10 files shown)


tools/irc/irc.php
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/php -q
```
XaeroDegreaz, line 1: Why remove the shebang?

XaeroDegreaz: Nevermind, saw the commit message

XaeroDegreaz (approved):
Good refactor; I like the fact that we can use PHPUnit instead of running those external scripts

PR #976, opened 2021-01-27
Remove port search reduction per evil good and incorporate in a base chance reduction to 4%

Page-:


hemberger (commented):
I'm okay with this simplification from a roleplaying perspective, because as I mentioned in Discord, I think it doesn't make sense for there to be _more_ Federal searches at low level ports. There is perhaps an argument that the search chance should be _lower_ at low level ports (because why would Federal authorities care about low level ports?), but I'm happy to motivate equality by imagining that the reduced presence of authorities at low level ports is balanced out by the obscurity in the crowd of the bustling high level ports.

From a gameplay perspective, it's important to recognize that this reduces the effectiveness of using low level ports to decrease alignment by a factor of 5 (since the search chance drops from 15% to 3%). It's probably a good idea to make this pathway to negative alignment less effective, because it's kind of a cheesy strategy, but we should still assess the impact of taking it away. Are there still effective alternatives for lowering alignment? Presumably only port raids and podding traders of races you're at peace with. Is that sufficient?

hemberger:
I meant to add that it's also a little weird that we have a base 3% chance reduction, and then the smuggling ship reduces that chance to -1%. Maybe it should be changed so the resulting chance is 0% instead, but obviously that doesn't matter from a gameplay perspective.

stupidnewbie:
Lowering alignment is pretty straightforward.  Just shoot low level ports.  If you can't find a race you're at peace with just use neutral ports.  It's faster and more turn efficient than careening around with contraband having to click every port.  Plus you gain credits looting the ports.

I've got more concern over losing the positive alignment gained from clean checks.  That extra few dozen points you get being checked while trading early on helps you stay green.  It's not a huge deal as long as you can constantly re-deputize and I won't rant about that loophole here.

It strikes me as more reasonable to set a higher check percent and then simply reduce it to 0 via smuggling ship bonus.  I'd guess something like 6% might be closer to realistic.  But I'm only guessing.  A few DB queries would probably shed better light on that. 

I do like the simplified calculation.  It's less crap to look up or remember.  And I do believe the smuggling craft bonus should be more substantial.  Right now it's good for 3 or 4 percent which is not a significant amount of the 1920 trade potential a Thief has.

PR #996, opened 2021-02-17
run-tests.yml: implement docker layer caching

hemberger:
This GitHub Action caches all layers that are created after the step
where it is specified (which is why we pre-fetch some pull-only
images that we know we'll be using to run the tests).

The goal is faster CI by speeding up the build stage.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -15,23 +15,26 @@ jobs:
           # Needed by Codecov
           fetch-depth: 2
 
-      # Pull the latest image to build, and avoid caching pull-only images.
-      # (docker pull is faster than caching in most cases.)
-      - name: Download pull-only images
-        run: docker-compose --env-file test/env pull mysql-integration-test flyway-integration-test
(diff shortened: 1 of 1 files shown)


hemberger:
This seems to be reducing the `build-and-test` time by about 40% (~90 seconds). While this is pretty good, it's perhaps not quite enough to warrant all the extra complexity. Furthermore, there are concerns about performance degradation over time as the cache fills up with obsolete build layers (see https://github.com/satackey/action-docker-layer-caching/issues/55).

For now, we'll stick with the simple, non-cached solution. But it's good to know that this works as expected should we have a more urgent need to reduce build times in the future. (It may also allow time for GitHub to provide a more "official" workflow for caching docker layers.)

First build, when the cache is empty:
image

Subsequent builds utilizing the cache:
image

hemberger:
There are more modern solutions to docker caching here: https://github.com/docker/build-push-action/tree/master/docs/advanced

hemberger:
Superseded by #1451.

PR #999, opened 2021-02-21
Update social logins

hemberger:
* Refactor into the PSR-4 autoloader.
* Switch Facebook provider to one that is PHP8 compatible.
* Add Google provider.


hemberger:
image

PR #1025, opened 2021-03-21
Replace custom php lint test with overtrue/phplint

hemberger:
This library runs the exact same linting check (i.e. `php -l`), except
it can be run in parallel. When using 8 cores, the check completes on
the entire repository in under 2 seconds (down from ~13 seconds).


hemberger:
Note that this is mostly useful for local testing, where we are more likely to have multiple cores.

PR #1034, opened 2021-04-05
Make SmrSession non-static

hemberger:
Static classes are an anti-pattern, and so we instead put the session
in the DI Container. When we want to access the session, we call
`SmrSession::getInstance`. (Note that this always returns the original
instance, unlike `MySqlDatabase::getInstance`, which returns a new
instance each call.)

Since the DI Container is responsible for constructing the SmrSession,
we no longer need to call `SmrSession::init` in SmrSession.class.php
outside the scope of the class definition.

Also add `Page::process` for handling the processing of engine files.


hemberger:
Superseded by #1038.

PR #1047, opened 2021-04-24
Change default date format to yyyy-mm-dd

hemberger:
Does not affect any existing accounts, only dates displayed when not
logged in and the default for new accounts.

This is a much more standard (and, in my opinion, readable) format
than the current default.

23/4/2021 -> 2021-04-23

Also changed the `account` table column names from `{date,time}_short`
to `{date,time}_format` for consistency with the SmrAccount method
name changes in #1046.


hemberger:
image

Issue #1183, opened 2021-12-09
Implement PHP 8.1 features :rocket: 


hemberger:
* [ ] Replace phpdoc `@return never` with native `never` return type
* [ ] Replace integer/string enumerations with native `enum` classes
* [ ] Use `readonly` properties in data storage classes with no setter methods

PR #1184, opened 2021-12-09
Use PHP 8.1 never return type

hemberger:
This replaces the `@return never` docstring used by some CI tools.

Related to #1183.


hemberger:
Note that Scrutinizer does not yet know about the `never` return type, because it does not (yet?) support PHP 8.1.

PR #1237, opened 2022-02-13
Miscellaneous updates

hemberger:
* Update D3, jQuery, and jQuery UI libraries
* Remove unused PHP code from header
* Fix port good spacing on Local Map


hemberger:
Old port good spacing (after whitespace trimming was removed):
image
New port good spacing:
image

PR #1263, opened 2022-03-07
Add Database::insert method

hemberger:
Manual construction of SQL statements is fragile and dangerous, since
there is no way to validate the resulting string apart from executing
the query.

We replace all basic INSERT statements with calls to Database::insert,
with structured data passed as an argument. (In the future, we should
apply this strategy to other statement types, but UPDATE statements,
as an example, are often more complex.)

This design has some very important benefits:

1. The key-value pairs are directly (and visually) associated, so it
   is much harder to pass the wrong value to a column.

2. It is impossible to create a string with a SQL syntax error, since
   all of the formatting (commas, parentheses, etc.) is automatic.

3. If we can unify the way that we escape content in the future, then
   we can construct the `$fields` array with raw values, and perform
   the escaping irrespective of the value types. (Though this could
   possibly degrade specificity, if we're not validating the types
   against a schema for the given table, for example.)

4. Just generally it is _much_ easier to read.

### DIFF
--- a/test/SmrTest/lib/DefaultGame/DatabaseIntegrationTest.php
+++ b/test/SmrTest/lib/DefaultGame/DatabaseIntegrationTest.php
@@ -248,7 +248,7 @@ class DatabaseIntegrationTest extends TestCase {
 		self::assertSame(0, $db->insert('debug', []));
 
 		// Non-zero insert ID when table has an auto-increment column
-		for ($i = 1; $i <=3; $i++) {
+		for ($i = 1; $i <= 3; $i++) {
 			self::assertSame($i, $db->insert('newsletter', []));
 		}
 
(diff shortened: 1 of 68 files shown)


test/SmrTest/lib/DefaultGame/DatabaseIntegrationTest.php
```diff
@@ -241,4 +241,26 @@ public function test_inversion_of_escape_and_get() : void {
...
 	}
 
+	public function test_insert() : void {
+		$db = Database::getInstance();
+
+		// Zero insert ID when table does not have an auto-increment column
+		self::assertSame(0, $db->insert('debug', []));
+
+		// Non-zero insert ID when table has an auto-increment column
+		for ($i = 1; $i <=3; $i++) {
```
hemberger, line 251:
```suggestion
		for ($i = 1; $i <= 3; $i++) {
```

PR #1407, opened 2022-08-07
Extend Rankings to game_stats.php

hemberger:
All remaining raw SQL queries are moved from the rankings_*.php pages
to the Rankings class. The game_stats.php page is re-implemented with
the Rankings class methods.

* Fix SQL error in alliance experience rankings in game_stats.php. In
  disbanded alliances (i.e. no players), we would be calling `SUM` on
  `NULL` due to a missing `COALESCE`, which was present in the
  (otherwise) identical query in rankings_alliance_experience.php
  in d9fca79617661.

* Fix bug calculating alliance kill counts in game_stats.php. We must
  get the count from the `alliance.alliance_kills` column instead of
  from the individual player kill counts. This way, only the kills
  awarded while the player was a member of the alliance will count
  toward the alliance total.

* The game_stats.php rankings automatically inherit the rank row
  styles from the Rankings class methods, so players will see their
  own trader/alliance highlighted.

### DIFF
--- a/src/engine/Default/game_stats.php
+++ b/src/engine/Default/game_stats.php
@@ -1,7 +1,10 @@
 <?php declare(strict_types=1);
 
+use Smr\Exceptions\PlayerNotFound;
+
 $template = Smr\Template::getInstance();
-$var = Smr\Session::getInstance()->getCurrentVar();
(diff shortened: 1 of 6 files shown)


hemberger:
image

PR #1474, opened 2022-11-22
SmrPlayer: simplify removeUnderAttack

hemberger:
Since the `player.under_attack` database column was added in #1020, this method does not need to (ab)use the session to cache its value. Instead, we can simply only change the value on non-ajax page loads.


hemberger:
This did not properly update "UnderAttack" if the attack occurred before an ajax update. This PR is superseded by 76e0b60b7657bcc23c20cfbad315514ca86d6b58.

PR #1553, opened 2023-02-12
Generate imprint dynamically from JSON

hemberger:
At MrSpock's request, imprint data is no longer hard-coded. Instead, we download a JSON file with up-to-date data.

Add the "lastUpdated" time to the display.

image


MichaelKunze:
LGTM

PR #1866, opened 2024-06-28
gameplay: optional permanent port destruction

hemberger:
* The `Game` class has a new property `destroyPorts` to track whether or not permanent destruction of ports is enabled. This setting can be modified when creating or editing a game.

* After a level 1 port is busted, it can optionally be destroyed if the feature is enabled for the game. No credits are gained, alignment is lost, and a news entry is created.

* Claiming a port for your race has been made mutually exclusive with all other post-bust options. It can only be done once, and you only get 50% of the credits.

* Changed terminology from "destroyed" to "busted" in Port/Planet interfaces to disambiguate between breached defenses and permanent destruction.

* Removed some hard-coded alignment change numbers in favor of named constants defined in `config.php`.

Screenshot from 2024-06-28 11-01-02


src/pages/Player/AttackPort.php
```diff
@@ -22,6 +22,10 @@ public function __construct(
 	}
 
 	public function build(AbstractPlayer $player, Template $template): void {
+		$sector = $player->getSector();
+		if (!$sector->hasPort()) {
+			(new CurrentSector(message: 'The port no longer exists!'))->go();
+		}
 		$port = $player->getSector()->getPort();
```
hemberger, line 29:
```suggestion
		$port = $sector->getPort();
```

PR #1967, opened 2024-12-03
Handle overflow in anon bank deposits

hemberger:
When depositing into an anon account, the database would overflow if the amount in the account would exceed MAX_MONEY. Now, the deposit will be adjusted so that it doesn't exceed MAX_MONEY, and if the account is already at MAX_MONEY, it will emit a comprehensible error.

### DIFF
--- a/src/pages/Player/Bank/AnonBankDetailProcessor.php
+++ b/src/pages/Player/Bank/AnonBankDetailProcessor.php
@@ -36,13 +36,23 @@ class AnonBankDetailProcessor extends PlayerPageProcessor {
 		]);
 		$trans_id = $dbResult->record()->getInt('max_id') + 1;
 
+		// Get the amount currently in this anon bank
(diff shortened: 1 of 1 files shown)


src/pages/Player/Bank/AnonBankDetailProcessor.php
```diff
@@ -36,25 +36,31 @@ public function build(AbstractPlayer $player): never {
...
+		]);
+		$anonAmount = $dbResult->record()->getInt('amount');
+
 		// Update the credit amounts for the player and the bank
 		if ($action === 'Deposit') {
 			if ($player->getCredits() < $amount) {
 				create_error('You don\'t own that much money!');
 			}
+			$amount = min($amount, MAX_MONEY - $anonAmount); // handle overflow
+			if ($amount == 0) {
```
hemberger, line 52:
```suggestion
			if ($amount === 0) {
```

PR #2043, opened 2025-03-17
Deduplicate damage taken display construction

hemberger:
The combat templates were all duplicating the exact logic for building the damage taken display HTML. We refactor this logic into the Template methods `display{Force}TakenDamage`.

Also fix the messages so that they are properly pluralized.

Note that the kamikaze effect is handled slightly differently. Now, if there is a kamikaze, it subtracts that from the number of mines that were destroyed by non-kamikaze. Previously, it only printed the mine damage if it was greater than kamikaze, but then displayed the total number of mines lost. I think the new logic is more correct, but from a practical perspective they are identical, because combat drones do not currently do both kamikaze and normal damage to mines.

### DIFF
--- a/src/lib/Smr/Force.php
+++ b/src/lib/Smr/Force.php
@@ -555,18 +555,18 @@ class Force {
 			}
 		}
 		return [
-						'KillingShot' => !$alreadyDead && !$this->exists(),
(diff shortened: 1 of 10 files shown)


hemberger:
image

PR #2091, opened 2025-05-23
Revamp the UniGen admin tool

hemberger:


hemberger:
Map selection options
image

Map auto-gen instead of manual/upload
image

Edit galaxy navigation options
image

Unreachable sectors sanity check
image

Galaxy insertion/deletion
image

View-only UniGen
image

hemberger:
No-Drop Zone location
image

PR #2211, opened 2025-10-03
Fix browser translation of input submit elements

hemberger:
When using an `<input type="submit" name=... value=... />` element, the "value" field is part of both the front-end display (the text displayed on the submit element) and the back-end processing (the data submitted with the request). As a result, when browsers automatically translate the page contents for non-English speakers, they change the value that is submitted with the request.

To fix this, we convert all named submit inputs to `<button>` elements instead, since these completely decouple the submitted data from the displayed text.

Add the `create_submit` function to standardize this process and ensure that we can modify this behavior universally if needed.

Also add a `create_submit_display` function for cases where there is only a single submit element in a form. Here we don't need the submit to supply a name-value pair in the submitted data, and the function input is used only for display text.

### DIFF
--- a/src/templates/Default/engine/Default/course_plot.php
+++ b/src/templates/Default/engine/Default/course_plot.php
@@ -23,7 +23,7 @@
 			<td><input required type="number" size="5" name="from" maxlength="5" class="center" value="<?php echo $ThisPlayer->getSectorID(); ?>"></td>
 			<td>&nbsp;&nbsp;&nbsp;&nbsp;To:&nbsp;</td>
 			<td><input required type="number" size="5" name="to" maxlength="5" class="center"></td>
-			<td>&nbsp;&nbsp;&nbsp;&nbsp;<?php echo create_submit('action', 'Plot course'); ?></td>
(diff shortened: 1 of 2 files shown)


src/templates/Default/engine/Default/course_plot.php
```diff
@@ -23,7 +23,7 @@
 			<td><input required type="number" size="5" name="from" maxlength="5" class="center" value="<?php echo $ThisPlayer->getSectorID(); ?>"></td>
 			<td>&nbsp;&nbsp;&nbsp;&nbsp;To:&nbsp;</td>
 			<td><input required type="number" size="5" name="to" maxlength="5" class="center"></td>
-			<td>&nbsp;&nbsp;&nbsp;&nbsp;<input type="submit" name="action" value="Plot Course"></td>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;<?php echo create_submit('action', 'Plot course'); ?></td>
```
hemberger, line 26:
```suggestion
			<td>&nbsp;&nbsp;&nbsp;&nbsp;<?php echo create_submit('action', 'Plot Course'); ?></td>
```

PR #2431, opened 2026-05-10
Bring Creonti and Ik'Thorne descriptions and images back in line with game lore

stupidnewbie:
[Creonti were originally](https://web.archive.org/web/19991023101931/http://merchant.shareplay.com/help/creonti.htm) described as a small creature and [Ik'Thorne were originally](https://web.archive.org/web/19991013124452/http://merchant.shareplay.com/help/ikthorne.htm) described as 11 foot tall humanoids.  The two have become crossed in the live game.


hemberger:
Thanks Newbie, but I would prefer to keep these as they are for now. I put a lot of effort into writing the new descriptions, and I think they are a significant improvement over the originals. Switching the Ikky and Creonti portraits felt more internally self-consistent to me (it seems natural for the 11-foot tall brutish creatures to have the largest and most powerful warships, and I always envisioned the Mothership as having a tiny bridge with its 50 plates of armor), and I don't feel like the lore suffers much as a result.

I'd be happy to continue discussing this issue, but I think any argument in favor of reversion will need to be motivated by more than just historical precedent, and, at the very least, will require dramatically improved racial descriptions.

stupidnewbie:
I really hadn't though about that before.  I probably could rewrite and dramatically improve racial descriptions.  However, I believe this suggestion, along with most of that reply, to be completely insincere and disingenuous.

PR #2517, opened 2026-08-24
Add missing "Under Attack" warning to Freon22 template

hemberger:
Users with the Freon22 template will now get the blinking warning when they are under attack.

Adjusted the CSS so that it lines up with the right ship panel.


hemberger:
image

